PR #194, opened 2026-05-12
Skip NuGet publish when NetPace.Core unchanged since previous tag

### LINKED ISSUE #186: Skip NuGet publish when NetPace.Core unchanged since previous tag
## Problem

Both [release-binaries.yml](.github/workflows/release-binaries.yml) and [publish-nuget.yml](.github/workflows/publish-nuget.yml) trigger on `push: tags: '*'`. Every tag therefore publishes a new `NetPace.Core` version to nuget.org — even when the only changes in the tag are CLI-side (`NetPace.Console`, docs, workflow, etc.).

`--skip-duplicate` doesn't help here: each tag is a new version number, so NuGet accepts the push as a new release. Result: `NetPace.Core` versions on nuget.org that are byte-equivalent to the previous version, eroding SemVer integrity for library consumers.

Approximately **90% of NetPace releases are CLI-only** — Core changes much more rarely. Optimising for that asymmetry is the goal.

## Proposed change

Keep the existing tag-driven model. Add a guard step at the top of `publish-nuget.yml`'s `build-and-publish` job that diffs `src/NetPace.Core/**` between the current tag and the previous tag. If unchanged, skip the pack + push steps. Binary releases (`release-binaries.yml`) are untouched.

This is the minimum-viable fix for the latent no-op-publish problem and preserves all current ergonomics:
- Tag-as-single-source-of-truth stays
- Hand-crafted GitHub Release notes stay (written for CLI users, as today)
- `<PackageReleaseNotes>` in [NetPace.Core.csproj:28](src/NetPace.Core/NetPace.Core.csproj#L28) (a static URL pointing to GitHub Releases) means NuGet consumers always click through to the same place — no per-version notes work needed
- When Core *does* change in a CLI release, it ships automatically with the same tag

## Implementation sketch

Edit [.github/workflows/publish-nuget.yml](.github/workflows/publish-nuget.yml). After the checkout step, add:

```yaml
      - name: Fetch tags for diff
        run: git fetch --tags --force --unshallow || git fetch --tags --force

      - name: Check if NetPace.Core changed since previous tag
        id: core_changed
        run: |
          set -euo pipefail
          PREV_TAG=$(git describe --tags --abbrev=0 "${GITHUB_REF}^" 2>/dev/null || echo "")
          if [ -z "$PREV_TAG" ]; then
            echo "No previous tag found — treating as first release, will publish."
            echo "changed=true" >> $GITHUB_OUTPUT
          elif git diff --quiet "$PREV_TAG"..."${GITHUB_REF}" -- src/NetPace.Core; then
            echo "NetPace.Core unchanged since $PREV_TAG — skipping NuGet pack/push."
            echo "changed=false" >> $GITHUB_OUTPUT
          else
            echo "NetPace.Core changed since $PREV_TAG — will publish."
            echo "changed=true" >> $GITHUB_OUTPUT
          fi
```

Then gate the existing `Pack` and `Publish to NuGet` steps with `if: steps.core_changed.outputs.changed == 'true'`. The `Restore dependencies` and `Build` steps can either also be gated (saves CI time on CLI-only releases) or left unconditional (acts as a smoke test that Core still builds). Recommend gating both — the binaries workflow already builds Core for the size-assertion check.

### Edge cases to handle
- **First-ever tag** (no previous tag): publish (handled above).
- **Re-tagging the same commit** (e.g. tag deletion + re-push): `git diff PREV...HEAD` is empty → skipped. Acceptable; matches current `--skip-duplicate` behaviour.
- **Annotated vs lightweight tags**: `git describe --tags --abbrev=0` works for both; verify in test.
- **Tag pushed without full history** (shallow clone): the `unshallow` fetch handles this.
- **Manual workflow re-run after a Core-only edit was hotfixed onto an existing tag**: the diff still reflects what's in the tag; if Core was changed and re-tagged, it will publish. Correct.

## Acceptance criteria

- [ ] `publish-nuget.yml` includes a `core_changed` step that compares `src/NetPace.Core/**` between the current and previous tags.
- [ ] `Pack` and `Publish to NuGet` steps are gated on `core_changed.outputs.changed == 'true'`.
- [ ] On a CLI-only tag, the workflow run shows the skip message and completes successfully without invoking `dotnet pack` or `nuget push`.
- [ ] On a tag that touches `src/NetPace.Core/**`, the workflow publishes as today.
- [ ] First-ever tag scenario publishes (no regression for the very first release).
- [ ] [release-binaries.yml](.github/workflows/release-binaries.yml) is **not** modified.

## Docs to update

Per the project memory rule (`feedback_release_pipeline_doc`), [docs/RELEASING.md](docs/RELEASING.md) must be updated alongside any release-pipeline change. Add a short section — suggested location after the existing release-matrix sections — covering:

- That NuGet publication is conditional on `src/NetPace.Core/**` having changed since the previous tag
- Why (avoid byte-equivalent NuGet versions, preserve SemVer meaning for library consumers)
- That the GitHub Release / binary-attachment flow is unchanged
- Note the consequence: `NetPace.Core` version numbers on nuget.org may skip values (e.g. `0.5.0` → `0.7.0`) when intermediate tags were CLI-only — this is intentional and informative

No README or USER_GUIDE changes needed (this is internal release tooling, not user-facing CLI behaviour).

## Out of scope

- Independent `<Version>` per project / csproj-driven release triggers — explored and rejected as overkill for the 90/10 asymmetry.
- Per-version NuGet release notes generated from PR titles — possible future enhancement, not needed now since [NetPace.Core.csproj:28](src/NetPace.Core/NetPace.Core.csproj#L28) points consumers to GitHub Releases.
- Splitting Console and Core into separate tag prefixes (e.g. `cli-x.y.z` / `core-x.y.z`) — explored and rejected for the same reason.
- Draft-release / two-workflow gating designs — explored and rejected as unnecessary ceremony.

## Context

Surfaced from a design conversation about decoupling CLI binary releases from NuGet publishing. Several alternative designs were considered (independent versioning, csproj-driven triggers, draft-release gating); the conversation settled on the minimal guard-step fix as the best match for NetPace's actual release cadence.

FrankRay78:
## Summary

Closes #186. Both `publish-nuget.yml` and `release-binaries.yml` trigger on every tag, so byte-equivalent `NetPace.Core` packages were being pushed to nuget.org on every CLI-only release (~90% of tags), eroding SemVer meaning for library consumers. This PR adds a tag-diff guard step to `publish-nuget.yml` that compares `src/NetPace.Core/**` between the current and previous tag and gates the Setup .NET / Restore / Build / Pack / Publish steps on the result. First-ever-tag returns `changed=true` (safe default); `release-binaries.yml` is untouched.

## Spec

N/A

## Changed Files

- .github/workflows/publish-nuget.yml
- docs/RELEASING.md

### DIFF
--- a/.github/workflows/publish-nuget.yml
+++ b/.github/workflows/publish-nuget.yml
@@ -21,11 +21,12 @@ jobs:
         id: core_changed
         run: |
           set -euo pipefail
-          PREV_TAG=$(git describe --tags --abbrev=0 "${GITHUB_REF}^" 2>/dev/null || echo "")
+          TAG_COMMIT=$(git rev-parse "${GITHUB_REF}^0")
+          PREV_TAG=$(git describe --tags --abbrev=0 "${TAG_COMMIT}^" 2>/dev/null || echo "")
           if [ -z "$PREV_TAG" ]; then
             echo "No previous tag found — treating as first release, will publish."
             echo "changed=true" >> $GITHUB_OUTPUT
-          elif git diff --quiet "$PREV_TAG"..."${GITHUB_REF}" -- src/NetPace.Core; then
+          elif git diff --quiet "$PREV_TAG".."${GITHUB_REF}" -- src/NetPace.Core; then
             echo "NetPace.Core unchanged since $PREV_TAG — skipping NuGet pack/push."
             echo "changed=false" >> $GITHUB_OUTPUT
           else
--- a/.github/workflows/publish-nuget.yml
+++ b/.github/workflows/publish-nuget.yml
@@ -13,7 +13,27 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Fetch tags for diff
+        run: git fetch --tags --force --unshallow || git fetch --tags --force
+
+      - name: Check if NetPace.Core changed since previous tag
+        id: core_changed
+        run: |
+          set -euo pipefail
+          PREV_TAG=$(git describe --tags --abbrev=0 "${GITHUB_REF}^" 2>/dev/null || echo "")
+          if [ -z "$PREV_TAG" ]; then
+            echo "No previous tag found — treating as first release, will publish."
+            echo "changed=true" >> $GITHUB_OUTPUT
+          elif git diff --quiet "$PREV_TAG"..."${GITHUB_REF}" -- src/NetPace.Core; then
+            echo "NetPace.Core unchanged since $PREV_TAG — skipping NuGet pack/push."
+            echo "changed=false" >> $GITHUB_OUTPUT
+          else
+            echo "NetPace.Core changed since $PREV_TAG — will publish."
+            echo "changed=true" >> $GITHUB_OUTPUT
+          fi
+
       - name: Setup .NET
+        if: steps.core_changed.outputs.changed == 'true'
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: 8.0.x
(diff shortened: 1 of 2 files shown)


FrankRay78:
@claude   A further commit has been pushed - can you code review this PR again?

FrankRay78:
@claude Please fix these which you identified in your code review (see: https://github.com/FrankRay78/NetPace/pull/194#issuecomment-4433215517):

<html>
<body>

Annotated-tag ^ bug — use git rev-parse "${GITHUB_REF}^0" first | Still needs fixing
-- | --
.. vs ... in git diff | Still needs fixing


</body>
</html>

PR #193, opened 2026-05-12
Win AOT Release

FrankRay78:
## Summary

Adds Windows Native AOT to the release matrix as two new archives — `netpace-{ver}-win-x64-aot.zip` and `netpace-{ver}-win-arm64-aot.zip` — single native `.exe`s with no .NET runtime dependency, bringing the per-release total to 16. Mirrors feature 001's Linux AOT pattern (`matrix.include:` entries + native runner per RID, `windows-latest` for x64 and `windows-11-arm` for arm64, since cross-OS AOT cross-compilation is fragile). Also tightens the archive-contents contract end-to-end: every release archive (all 16) must now contain exactly one entry, enforced by a new "Verify archive contents" workflow step; managed PDBs are embedded via `<DebugType>embedded</DebugType>` in both `NetPace.Console.csproj` and `NetPace.Core.csproj` so no side `.pdb` ships in any variant, and the Windows AOT native PDB is scrubbed before zipping. Includes a runner-OS guard step that fails fast on a `matrix.runs_on` typo (would otherwise silently fall back to `ubuntu-latest`).

## Spec

[specs/002-win-aot-release](specs/002-win-aot-release)

## Changed Files

- .claude/settings.json
- .github/workflows/release-binaries.yml
- .specify/feature.json
- CLAUDE.md
- README.md
- docs/RELEASING.md
- docs/change-intent-records/2026-05-02-linux-native-aot-release.md
- specs/002-win-aot-release/checklists/requirements.md
- specs/002-win-aot-release/contracts/release-matrix.md
- specs/002-win-aot-release/data-model.md
- specs/002-win-aot-release/plan.md
- specs/002-win-aot-release/quickstart.md
- specs/002-win-aot-release/research.md
- specs/002-win-aot-release/spec.md
- specs/002-win-aot-release/tasks.md
- specs/002-win-aot-release/test-plan.md
- src/NetPace.Console/NetPace.Console.csproj
- src/NetPace.Core/NetPace.Core.csproj

## New Artifacts

- Spec: [specs/002-win-aot-release](specs/002-win-aot-release)

PR #198, opened 2026-05-13
Archive release binary by name instead of scrubbing sidecars

FrankRay78:
## Summary

- Replaces the extension-based scrub approach (e.g. `rm *.pdb`) in the **Create archive** step with a whitelist: archive the binary by name (`NetPace.exe` / `NetPace`).
- Fixes the underlying cause of the `linux-x64-aot` and `linux-arm64-aot` failures on tag `0.23.2` (stray `NetPace.dbg`) without adding another `rm`-by-extension band-aid that would be due to break again on the next toolchain artefact.
- Updates `docs/RELEASING.md` §Archive-contents contract to describe the whitelist mechanism and call out the by-products it now excludes by construction (`NetPace.Core.xml`, `NetPace.pdb`, `NetPace.dbg`).

## Why this over the targeted `rm -f ./*.dbg` fix

The previous design tried to satisfy "archive contains exactly one entry" through three layered mechanisms (XML doc suppression, `DebugType=embedded`, per-extension `rm` in the workflow). Each new toolchain by-product becomes a new whack-a-mole hit — most recently `NetPace.dbg` on Linux AOT. Whitelisting the binary by name collapses all three to a single concept and immunises the release pipeline against whatever the next toolchain version decides to emit alongside the binary.

The **Verify archive contents** step is retained as a safety net — it now also catches a silent build that produced no binary at all.

## Test plan

- [x] `dotnet build src` clean, 0 warnings, 0 errors locally
- [ ] Push a throwaway tag (e.g. `test-archive-whitelist-0.0.1`) to dry-run all 16 matrix jobs without creating a real GitHub Release; confirm "Verify archive contents" passes for every variant including the previously-broken `linux-x64-aot` and `linux-arm64-aot`; then delete the test tag
- [ ] After merge, the next real release tag should produce 16 archives each containing exactly one entry

Closes the regression first surfaced on tag 0.23.2.

---

nb. This PR attempts to fix the mess caused by: https://github.com/FrankRay78/NetPace/pull/193

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -131,20 +131,22 @@ jobs:
             -p:FileVersion=${{ steps.get_version.outputs.version }} \
             -p:InformationalVersion=${{ steps.get_version.outputs.version }}
 
-      # shell: bash forces Git Bash on the Windows runners so the same script (zip vs tar, glob branching) runs uniformly across all matrix entries.
+      # shell: bash forces Git Bash on the Windows runners so the same script (zip vs tar) runs uniformly across all matrix entries.
+      # Whitelist approach: archive the executable by name rather than the entire publish directory. All variants
+      # (PublishSingleFile and Native AOT) produce a single binary by design; the publish dir may also contain
+      # toolchain by-products (NetPace.Core.xml, NetPace.pdb on Windows AOT, NetPace.dbg on Linux AOT) that
+      # must never ship to end users. Archiving by name keeps the release contract immune to whatever the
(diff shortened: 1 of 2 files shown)


FrankRay78:
@claude 

> Run set -euo pipefail
D:\a\_temp\6c724507-5db4-4053-a3d0-d3c78a8649f5.sh: line 5: zip: command not found
Error: Process completed with exit code 127.

https://github.com/FrankRay78/NetPace/actions/runs/25800045766/job/75787268564

PR #213, opened 2026-07-09
Update Slash Commands

FrankRay78:
Ports a set of agentic-development improvements from the companion repo into NetPace, adapting everything IMS-specific to this repo. Three independent themes.

## What changes

- **Flatten `.claude/memory/`.** All 14 memory entries move out of the `process/` and `project/` subdirectories into a single flat directory, and `MEMORY.md` becomes one uncategorised list. The two-bucket taxonomy was a redundant third encoding — kind is already carried by the `feedback_`/`project_` filename prefix and the `type:` frontmatter — and the flat shape matches the companion repo so shared tooling copies over without per-repo path rewrites.
- **Upgrade `/capture-learnings`.** Adds a triage step that prefers a deterministic mechanism (a hook, test, or `settings.json` change) over a memory entry wherever a learning is mechanically enforceable, falling back to memory only for judgment or context no mechanism can capture. Deterministic changes are propose-then-apply — never a silent `settings.json` edit.
- **Add the `/diagnose` skill.** A project-agnostic `SKILL.md` (reproduce → build a feedback loop → hypothesise → instrument → fix → regression-test → post-mortem) plus a NetPace-specific `NETPACE.md` sidecar of concrete feedback-loop builders: single-test `--filter`, Verify snapshot loop, MockHttp payload replay, the local Docker OoklaServer, and a differential worktree.

- **The memory moves are pure `R` renames** (content unchanged) except `MEMORY.md`, whose links are re-pointed to the flat paths. No memory entry cross-references another by path, so nothing else needed updating.
- **`SKILL.md` is intentionally left generic** — it keeps the shared reference to `scripts/hitl-loop.template.sh` (which doesn't exist here) rather than forking the shared skill body. The concrete, NetPace-only material lives in `NETPACE.md`.
- **The three re-pointed references in `capture-learnings.md`** replace IMS paths that don't exist here (`docs/TESTING.md`, an IMS memory file) with NetPace's actual TDD/hooks reality and the constitution's Testing Standards.
- `NETPACE.md`'s factual claims (ports, endpoints, test-infra names, file paths, the "no real-network test category" assertion) were fact-checked against the live repo; a few inaccuracies caught in review are already fixed in this branch.

## How to verify

- [ ] `ls .claude/memory/` — flat, no `process/`/`project/` subdirs; every `MEMORY.md` pointer resolves and every file has exactly one pointer.
- [ ] `/capture-learnings` and `/diagnose` appear as available slash commands/skills.
- [ ] Spot-check `NETPACE.md`: `./docker/ooklaserver/start.sh` (port 18080), and a listed `dotnet test src --filter ...` command resolves to a real test.


FrankRay78:
- Bugs or correctness issues

PR #214, opened 2026-07-10
Upgrade spec-kit to 0.12.10 and switch scripts to bash

FrankRay78:
## Why

Development moved from Windows to Linux, where there's no `pwsh` — but the core spec-kit automation scripts were PowerShell-only, so the SDD skills couldn't run. Investigation confirmed the `.specify/` toolkit is **stock `github/spec-kit`** (not a fork; its `0.8.x`/`0.12.x` version line is unrelated to NetPace releases), so this takes the sanctioned route: upgrade to the latest release via the `specify` CLI and switch the script variant to bash.

## What changes

- Upgraded the spec-kit toolkit **0.8.12.dev0 → 0.12.10** in place (`specify init --here --force --integration claude --script sh`).
- Core automation is now **bash** (`.specify/scripts/bash/*.sh`); skills invoke `*.sh --json` instead of `*.ps1 -Json`.
- Removed the now-orphaned PowerShell core scripts and the dropped `update-context.*` integration scripts, plus their manifest entries.
- Repointed the `.claude/settings.json` allowlist from `powershell`/`pwsh` to the bash invocations.
- Picked up the new `speckit-converge` skill that ships with 0.12; added a memory note documenting the upgrade procedure.
- Removed the leftover `specs/003-profile-cli-switch` spec folder (merged feature; specs aren't kept on main).

## Non-obvious things for the reviewer

- **Customizations were preserved by design** — the upgrade is hash-guarded via `speckit.manifest.json` (overwrites only unmodified base files). The testplan extension, `templates/overrides/spec-template.md`, the constitution, and custom `speckit.*` commands are untouched.
- **Behavior change from 0.8→0.12:** `get_current_branch` is now decoupled from git — feature resolution is driven by `.specify/feature.json` / the feature-dir basename, not the checked-out branch. So `BRANCH` in script output is the feature name, not the git branch.
- The core bash scripts are **byte-identical** to a clean canonical 0.12.10 install (verified by diffing against a throwaway `specify init`).
- The git-extension bash scripts show only **mode changes** (executable bit set by init), no content changes.
- The git extension intentionally keeps **both** bash and PowerShell variants — those remaining `.ps1` references are its dual-variant design, not leftovers.

## How to verify

- [ ] `bash .specify/scripts/bash/check-prerequisites.sh --json --paths-only` returns valid JSON (exit 0).
- [ ] Skills reference bash: `grep -r "scripts/bash" .claude/skills/speckit-*/SKILL.md` (no `scripts/powershell`).
[...]

PR #215, opened 2026-07-10
Guard Speckit Files

FrankRay78:
## Why

Upstream-managed spec-kit files (core templates, bash scripts, the SDD `SKILL.md` skills) carry small local customizations layered on top of the upstream template. A headless agent — or an `specify init --force` regen — can silently overwrite those files with no gate, which is exactly how the `disable-model-invocation: true → false` flip slipped into #214. Nothing in the harness stopped it: `.claude/settings.json` blanket-allows `Edit(**)`/`Write(**)` with `defaultMode: acceptEdits`.

This adds a hard edit-time gate on the files that are **not** designed as extension points.

## What changes

`permissions.deny` rules in `.claude/settings.json` block `Edit`/`Write`/`MultiEdit` on:

- `.claude/skills/speckit-*/SKILL.md` — the 10 SDD skills tracked in `claude.manifest.json` (glob also covers the 5 `speckit-git-*` extension skills, a harmless superset)
- `.specify/templates/*.md` — the 6 core templates
- `.specify/scripts/bash/*.sh` — the 5 core scripts

`deny` overrides the existing blanket allow, so an agent is blocked *before* the edit lands, not at commit time.

- **The globs deliberately match the two spec-kit manifests' file lists.** `.specify/templates/*.md` and `.specify/scripts/bash/*.sh` are single-level globs (`*` doesn't cross `/`), so genuine extension points stay editable: `.specify/templates/overrides/**`, `.specify/extensions/**`, `constitution.md`, and `feature.json` are untouched by the deny.
- **`speckit-*` over the manifest's exact 10 is intentional.** A glob survives new SDD skills being added (e.g. `speckit-converge`) and can't silently fail open the way a clever character-class exclusion could. Protecting the git-extension skills too is consistent with the goal.
- **Known side door:** this gates the `Edit`/`Write`/`MultiEdit` tool path only. A `sed -i` / `git checkout` via Bash could still mutate these files — out of scope here; a CI content-invariant check would be the backstop if that's ever needed.
- **`deny`, not `ask`, by choice:** a future legitimate hand-edit requires temporarily removing the relevant rule. Chosen over `ask` for a firmer guarantee against headless drift.

## How to verify

- [ ] `jq empty .claude/settings.json` — valid JSON
- [ ] Ask an agent to edit `.claude/skills/speckit-plan/SKILL.md` or `.specify/scripts/bash/common.sh` — the `Edit`/`Write` call is denied
- [ ] Confirm `.specify/templates/overrides/spec-template.md` and `.specify/memory/constitution.md` remain editable (not matched by the single-level globs)

## Related

- Follow-up to #214 (the spec-kit 0.12.10 upgrade whose regen caused the `disable-model-invocation` regression)

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -81,7 +81,16 @@
       "Read(**/secrets.*)",
       "Read(**/.ssh/**)",
       "Read(**/appsettings.*.json)",
-      "Read(**/appsettings.json)"
+      "Read(**/appsettings.json)",
+      "Edit(.claude/skills/speckit-*/SKILL.md)",
+      "Write(.claude/skills/speckit-*/SKILL.md)",
+      "MultiEdit(.claude/skills/speckit-*/SKILL.md)",
+      "Edit(.specify/templates/*.md)",
+      "Write(.specify/templates/*.md)",
+      "MultiEdit(.specify/templates/*.md)",
+      "Edit(.specify/scripts/bash/*.sh)",
+      "Write(.specify/scripts/bash/*.sh)",
+      "MultiEdit(.specify/scripts/bash/*.sh)"
     ],
     "ask": [
       "Bash(rm:*)",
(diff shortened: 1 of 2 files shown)


FrankRay78:
- Bugs or correctness issues

PR #216, opened 2026-07-10
Agentic Harness Updates

FrankRay78:
## Why

A batch of five discrete, independent hardening improvements to NetPace's agentic-development harness, porting practices already proven on a sibling .NET/spec-kit repo (IMS). Each addresses a specific gap: duplicated policy, missing collaboration guidance, an un-captured lesson, and — the substantive one — no enforcement against silently skipped tests.

## What changes

- **Constitution de-duplication** — removed the "no failing tests / build warnings" rule from `Development Workflow > Git Workflow`; it now lives once in CLAUDE.md's actionable "Working with Claude Code" list (constitution `1.3.0 → 1.3.1`).
- **New CLAUDE.md collaboration rules** — no hard-wrapped markdown prose, frame decisions in plain-language consequences before mechanism, and don't fold an unrelated mission into an in-flight branch.
- **New memory entry** — read the source that implements a subsystem before designing a fix that depends on its behaviour.
- **Constitution Principle X — No Skipped Tests (NON-NEGOTIABLE)** — bans the entire xUnit skip family (constitution `1.3.1 → 1.4.0`).
- **Gate hook** — `.claude/hooks/no-skipped-tests.sh` blocks any `git commit` that would introduce a skip-family construct under `src/`, wired first into `settings.json`'s `PreToolUse` chain, plus a `--check` mode for CI/manual scans.

- **The hook fails *closed*, but only after classifying a command as `git commit`.** Every non-commit Bash call is waved through before any jq/`src/` dependency check runs — this is deliberate, so a box without `jq` can't deadlock on unrelated commands (even `apt install jq`). For an actual commit, a missing `jq`, an unlocatable `src/`, or a scan error blocks rather than silently allowing.
- **Executable bit was set via `git update-index --chmod=+x`, not `chmod`.** `Bash(chmod:*)` is in this repo's `settings.json` deny list, so the filesystem `chmod` is blocked; the committed file mode is `100755` regardless.
- **`NETPACE_ALLOW_SKIPS=1` is an announced escape hatch** — it prints a warning on every invocation so it can never be silently in effect. The header comment marks it for deletion once the gate is trusted.
- **The wired hook only arms on the next session** — harness settings load at session start, so it does not gate commits made in the session that added it. It was validated standalone instead (see below).
- No production `.cs` was touched — this PR is docs, memory, constitution, one shell hook, and settings wiring only.

## How to verify

- [ ] `./.claude/hooks/no-skipped-tests.sh --check` exits `0` on the clean tree.
- [ ] Piping a `git commit` hook payload with no skip present exits `0`; adding a probe line containing `[Fact(Skip = "temp")]` under `src/` and re-piping exits `2` with a `BLOCKED:` message naming Principle X.
- [ ] `settings.json` remains valid JSON and the new entry is first in `PreToolUse[0].hooks`.
- [ ] `dotnet build ./src` is clean (0 warnings, 0 errors). Note: the test *suite* targets .NET 8.0; run it on a machine with the 8.0 runtime installed.

### DIFF
--- a/.claude/hooks/no-skipped-tests.sh
+++ b/.claude/hooks/no-skipped-tests.sh
@@ -1,0 +1,133 @@
+#!/usr/bin/env bash
+#
+# no-skipped-tests.sh — static gate banning the entire skipped-test family (Constitution Principle X).
+#
+# A skipped test reports "green" while checking nothing, so it is silent non-coverage.
+# This gate blocks any `git commit` while a banned construct exists anywhere under src/.
+# Legitimate needs are met without skips: fail loudly on a missing dependency, exclude
+# destructive opt-in suites by [Trait("Category", …)], or document a genuinely untestable
+# branch with a comment at the site.
+#
+# Two modes:
+#   --check        Scan the tree and exit 1 if any banned construct is found (for CI / manual).
+#   (default)      PreToolUse hook: read hook JSON on stdin, block (exit 2) a `git commit`
+#                  that would carry a banned construct.
+#
+# For a commit it is gating, the gate fails CLOSED: if it cannot locate src/, cannot scan, or
+# cannot parse its hook payload (e.g. jq missing), it blocks rather than silently allowing — a
+# gate that fails open is the same silent non-coverage it exists to prevent. These fail-closed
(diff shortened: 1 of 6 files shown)


FrankRay78:
- Bugs or correctness issues

PR #233, opened 2026-07-30
Permission Rule Warnings

FrankRay78:
## Why

Every session opened with twelve permission warnings:

```
Permission deny rule "MultiEdit(.specify/templates/*.md)" matches no known tool — check for typos.
Permission allow rule (.claude/settings.json): Write(**) is not matched by file permission checks —
only Edit(path) rules are. Use Edit(**) instead (Edit rules cover all file-editing tools).
```

File-permission checks consult `Edit(path)` rules only, and an `Edit` rule covers **every** file-editing tool — Write included. `MultiEdit` is a legacy tool name that matches nothing at all. So the parallel `Write(...)`/`MultiEdit(...)` deny rules added alongside each guarded spec-kit path in #215 denied nothing, and the `Glob(**)`/`MultiEdit(**)`/`Write(**)` allow entries allowed nothing already covered by `Read(**)`/`Edit(**)`. Twelve inert lines, twelve warnings.

## What changes

- `settings.json` `deny` — back to one `Edit(path)` rule per guarded spec-kit path. The guard is unchanged in strength.
- `settings.json` `allow` — dropped the three inert entries.
- The two docs that described the guard as covering "Edit/Write/MultiEdit" now describe it accurately, and the CIR carries a dated Decision 4 recording that the parallel rules denied nothing and must not be re-added.
- New memory entry `speckit_file_guard.md` — NetPace had none, so nothing was stopping this being "strengthened" a third time.

Permission rules and hook `matcher` fields are different mechanisms. A hook matcher matches tool names directly, so a `Write`/`MultiEdit` arm in a `"matcher"` string is a separate question from the above — NetPace's only hook matcher is `Bash`, so nothing here touches hooks, and the memory entry says so explicitly to stop a future reader over-applying the rule.

## How to verify

- [ ] Start a session on this branch — the twelve warnings are gone.
- [ ] Ask an agent to `Write` to a path under `.specify/templates/` — still denied with only the `Edit(...)` rule present. (Verified during authoring: a `Write` to `.specify/templates/deny-probe.md` was blocked.)
- [ ] Confirm a genuine extension point (`.specify/templates/overrides/**`, `.specify/memory/constitution.md`) is still editable.

## Related

- Amends #215 (Guard Speckit Files).

### DIFF
--- a/.claude/memory/MEMORY.md
+++ b/.claude/memory/MEMORY.md
@@ -14,6 +14,7 @@
 - [NetPace CLI feature issues must scope user-facing docs from the start](feedback_cli_feature_doc_scope.md) — first draft must include README --help, USER_GUIDE, design-doc cross-refs (no CHANGELOG — release notes are GitHub-auto-generated)
 - [Release-pipeline changes must update docs/RELEASING.md](feedback_release_pipeline_doc.md) — release matrix, runner-per-RID, naming convention, smoke-test contract live there; out-of-sync = future RID work costs extra
 - [Verify-snapshot tests count as coverage in NetPace.Console.Tests](feedback_console_output_snapshot_coverage.md) — check Expectations/*.verified.txt before reporting an output mode as untested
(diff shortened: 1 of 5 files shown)


FrankRay78:
- Bugs or correctness issues

PR #162, opened 2026-04-13
SDD Agentic Development Workflow

FrankRay78:
## Summary

This branch installs and configures the Spec-Kit agentic development workflow for the NetPace repository. It adds Claude Code slash commands and skills covering the full spec-driven cycle (specify, clarify, plan, implement, test-checklist, and git operations), configures a GitHub Actions workflow for automated Claude-assisted PR review, and establishes convention documentation for C# style and change intent records. A follow-up set of PR review fixes corrects branch references (`master` → `main`), spec directory paths (`.specify/specs/` → `specs/`), GitHub Actions permissions, and manifest completeness for the git extension skills.

## Spec

N/A

## Changed Files

.claude/CLAUDE.md
.claude/commands/audit-deadcode.md
.claude/commands/raise-pr.md
.claude/commands/review-slop.md
.claude/commands/speckit.testchecklist.md
.claude/commands/speckit.testplan.md
.claude/settings.json
.claude/skills/speckit-analyze/SKILL.md
.claude/skills/speckit-checklist/SKILL.md
.claude/skills/speckit-clarify/SKILL.md
.claude/skills/speckit-constitution/SKILL.md
.claude/skills/speckit-git-commit/SKILL.md
.claude/skills/speckit-git-feature/SKILL.md
.claude/skills/speckit-git-initialize/SKILL.md
.claude/skills/speckit-git-remote/SKILL.md
.claude/skills/speckit-git-validate/SKILL.md
.claude/skills/speckit-implement/SKILL.md
.claude/skills/speckit-plan/SKILL.md
.claude/skills/speckit-specify/SKILL.md
.claude/skills/speckit-tasks/SKILL.md
.claude/skills/speckit-taskstoissues/SKILL.md
.github/pull_request_template.md
.github/workflows/claude.yml
.specify/extensions.yml
.specify/extensions/.registry
.specify/extensions/git/README.md
.specify/extensions/git/commands/speckit.git.commit.md
.specify/extensions/git/commands/speckit.git.feature.md
.specify/extensions/git/commands/speckit.git.initialize.md
.specify/extensions/git/commands/speckit.git.remote.md
.specify/extensions/git/commands/speckit.git.validate.md
.specify/extensions/git/config-template.yml
.specify/extensions/git/extension.yml
.specify/extensions/git/git-config.yml
.specify/extensions/git/scripts/bash/auto-commit.sh
.specify/extensions/git/scripts/bash/create-new-feature.sh
[...]

PR #236, opened 2026-08-27
Move formatting to ship cadence

### LINKED ISSUE #234: Reconcile format-on-commit against the guide's ship-cadence recommendation
## The divergence

`docs/agentic-workflow.md` (merged in #232) added a section arguing that formatting does not belong on the inner loop:

> Formatting is cosmetic, and cosmetic work does not belong on the inner loop. A format-on-commit hook taxes **every** commit — on a real codebase the tool's workspace load is measured in tens of seconds — to fix something no reviewer would have caught anyway. Fold the format step into the ship flow instead, where it runs once per PR at a cadence that already costs minutes.

NetPace does the opposite. `.claude/settings.json` wires a `PreToolUse(Bash)` hook that fires on `git commit`:

```
STAGED=$(git diff --cached --name-only | grep "[.]cs$"); [ -z "$STAGED" ] && exit 0; INCLUDE=$(echo "$STAGED" | tr '\n' ' '); dotnet format style --include $INCLUDE && dotnet format whitespace --include $INCLUDE && echo "$STAGED" | xargs -d '\n' git add
```

`docs/agentic-workflow-NetPace.md` records this under **"Divergence to reconcile — format-on-commit"** and deliberately leaves the call open:

> This is an **unreconciled divergence** — either move formatting into the ship flow, or record here why per-commit is deliberately kept. Flagged, not silently retained.

Split out of #230, which delivered the guide itself. #230 asserted "nothing else in NetPace needs to change"; this is the one place that turned out not to hold.

## The decision

Two admissible outcomes; either one closes this issue.

**A. Move formatting to ship cadence.** Drop the `PreToolUse` hook and add a `dotnet format` step to `/ship` ahead of the suite run, so formatting changes are covered by the test gate rather than landing after it. Removes a per-commit tax that buys nothing a reviewer would have caught. Cost: an agent working outside `/ship` leaves unformatted code on the branch until ship time.

**B. Keep per-commit, and write down why.** Record the justification in `docs/agentic-workflow-NetPace.md` in place of the current "to reconcile" wording. Legitimate grounds would be evidence that the tax is small here — `src/` is a single small solution, so the guide's "tens of seconds" figure may not hold — or that formatting at ship time produces noisy diffs against a review already in flight.

## Suggested first step

Measure before choosing. Time `dotnet format style --include` plus `dotnet format whitespace --include` over a representative staged set on this repo. If the per-commit cost is genuinely negligible, B is defensible and cheap; if it is the guide's tens of seconds, A wins on the guide's own argument.

Whichever is chosen, update `docs/agentic-workflow-NetPace.md` so the divergence section states a settled position rather than an open one.

FrankRay78:
Closes #234.

## Why

`docs/agentic-workflow.md` argues formatting does not belong on the inner loop — it should run once per PR, at a cadence that already costs minutes. NetPace did the opposite via a `PreToolUse(Bash)` hook on `git commit`, and the delta doc flagged that as an **unreconciled divergence**. This settles it in the guide's favour, on evidence.

Two measurements decided it:

- **The per-commit cost was real.** Timed against `src/NetPace.sln`: **21.4s** for a single staged file, **29.5s** for a seven-file set. The cost is MSBuild *workspace load*, not file count, so staging fewer files does not help. The guide's "tens of seconds" holds even on a solution this small — which was the main argument for keeping per-commit.
- **The value was near zero.** The hook had in fact been a **no-op since #162**. It invoked `dotnet format` with no workspace argument, and `dotnet format` searches only the current directory — where NetPace has no `.sln`. Every invocation died with `Could not find a MSBuild project file or solution file` and exit 1, a *non-blocking* `PreToolUse` result, so commits went through unformatted for four and a half months. Total drift over that window: **3** import-ordering findings and **223** trailing-whitespace fixes.

## What changes

- **The format hook is deleted.** The other three hooks are untouched.
- **`/ship` step 1 splits into 1a (format) and 1b (full test run).** Step 1a runs `dotnet format style/whitespace ./src/NetPace.sln` and commits any result on its own *before* the suite.
- **`*.cs text eol=lf`** added to `.gitattributes` (separate commit).
- **The delta doc's "Divergence to reconcile" section becomes a settled position**, and the gates table moves formatting from a hook to a `/ship` step.

- **The `./src/NetPace.sln` argument in step 1a is load-bearing, not decorative.** Omitting it is precisely the bug that made the old hook a silent no-op. Please don't "tidy" it away.
- **Step 1a commits its own changes before the suite runs.** That is deliberate: step 0 requires a clean tree so that step 3 can assume "anything in the tree after review is a review edit". Formatting must restore that invariant rather than leaving edits for the review commit to sweep up.
- **The `.gitattributes` change is zero-diff.** The index was already LF; only Windows working trees were CRLF, via `core.autocrlf=true`. No renormalisation commit is needed. Windows trees created before this line need a one-time refresh to pick the attribute up.
- **No source files are touched by this PR.** The first `/ship` run on any branch will produce a one-off repo-wide `style: apply dotnet format` commit (the 3 + 223 findings above). That is expected, and is better done deliberately on its own branch than folded in here.
- **IMS is deliberately not synced.** The two `ship.md` files are otherwise the same document with per-repo slot substitutions, so this does open a divergence in it. IMS is code-complete; the delta doc records how step 1a ports when it next takes a version (behind IMS's existing source-root probe, plus `gofmt` for chaincode).

## How to verify

- [ ] `dotnet format style ./src/NetPace.sln --verify-no-changes` reports the 3 import-ordering findings — i.e. the command in step 1a actually resolves a workspace, unlike the deleted hook.
- [ ] Running the same command *without* the `./src/NetPace.sln` argument from the repo root reproduces the original `Could not find a MSBuild project file or solution file` failure.
- [ ] `git ls-files --eol src/NetPace.Core/SpeedScale.cs` shows `attr/text eol=lf`, and `git status` shows no source files modified.
- [ ] `git commit` no longer triggers a "Formatting staged .cs files..." hook.

## Related

- #234 — the reconciliation issue, including the full investigation and measurements.
- #232 — added the guide section this reconciles against.
- #162 — introduced the hook.

### DIFF
--- a/.claude/commands/ship.md
+++ b/.claude/commands/ship.md
@@ -27,7 +27,7 @@ Read `CLAUDE.md` for project context before proceeding.
    dotnet format style ./src/NetPace.sln && dotnet format whitespace ./src/NetPace.sln
    ```
 
-   The explicit `./src/NetPace.sln` argument is **required, not decorative**: `dotnet format` looks for a project or solution in the *current directory only*, and NetPace's solution is at `src/`, not the repo root. Omitting it fails with `Could not find a MSBuild project file or solution file` — which is exactly how the retired format-on-commit hook silently no-op'd for four months (#234).
+   The explicit `./src/NetPace.sln` argument is **required, not decorative**: `dotnet format` looks for a project or solution in the *current directory only*, and NetPace's solution is at `src/`, not the repo root. Omitting it fails with `Could not find a MSBuild project file or solution file`.
 
    - Formatting runs **once per ship**, not per commit. It is cosmetic work at a cadence that already costs minutes — see *Formatting is not verification* in [agentic-workflow.md](../../docs/agentic-workflow.md).
    - **If formatting changed files, commit them now** — `git add -A` and a `style: apply dotnet format` message — *before* running the suite. This restores the clean working tree step 0 established, which is what keeps step 3's "anything in the tree is a review edit" invariant true. Do not carry format edits forward into the review commit; they are a separate concern and belong in their own commit.
--- a/.claude/commands/ship.md
+++ b/.claude/commands/ship.md
@@ -4,7 +4,7 @@ description: Orchestrator that turns "implementation looks done" into a reviewed
 
 Read `CLAUDE.md` for project context before proceeding.
 
-`/ship` composes NetPace's existing commands behind one hard gate: **the full test suite runs first, and nothing downstream happens unless it is green.** The gate is structural — the review step is downstream of the step-1 exit code, so it cannot begin against un-verified or red code. Do not add a hook to police this ordering; the exit code *is* the gate.
+`/ship` composes NetPace's existing commands behind one hard gate: **the full test suite gates everything downstream, and no review or PR happens unless it is green.** The gate is structural — the review step is downstream of the step-1 exit code, so it cannot begin against un-verified or red code. Do not add a hook to police this ordering; the exit code *is* the gate. The one step that precedes the suite is formatting (step 1a), which is cosmetic and is itself covered by the gate that follows it.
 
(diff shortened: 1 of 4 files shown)


FrankRay78:
- Bugs or correctness issues

FrankRay78:

- Bugs or correctness issues

Issue #206, opened 2026-06-21
some servers show an upload speed of 0 bps, unlike the speetest.net website


helptiger:
### Steps to reproduce
--no-download --no-latency --server "http://ffm.wsqm.telekom-dienste.de:8080/speedtest/upload.php"
--no-download --no-latency --server "http://ham.wsqm.telekom-dienste.de:8080/speedtest/upload.php"
--no-download --no-latency --server "http://mue.wsqm.telekom-dienste.de:8080/speedtest/upload.php"

### Additional context
The problem has been occurring regularly since June 6, 2026. Other Deutsche Telekom servers are providing correct readings.
--no-download --no-latency --server "http://ber.wsqm.telekom-dienste.de:8080/speedtest/upload.php"
The app and NuGet library are affected

Thanks for the support!

FrankRay78:
I'm happy to take a look into this, a little busy ATM but I will make some time this/next week, all going well. Thanks for reporting.

helptiger:
Here are some other servers that show 0.0 Mbps in the upload speed test:

- Plusnet GmbH, Düsseldorf
"http://speedtest.dus.plusnet.de:8080/speedtest/upload.php"

- Hivelocity, Frankfurt
"http://speedtest.fra1.hivelocity.net:8080/speedtest/upload.php"

- WEMACOM Telekommunikation GmbH, Frankfurt
"http://de-ffm1-ookla-01.wemacom.net:8080/speedtest/upload.php"

- suec//dacor GmbH, Frankfurt
"http://speedtest1.dacor.de:8080/speedtest/upload.php"?

- Digi Turunc, Amsterdam
"http://speedtest-nl.digiturunc.com:8080/speedtest/upload.php"

- Clouvider Ltd, Amsterdam
"http://ams.speedtest.clouvider.net:8080/speedtest/upload.php"

FrankRay78:
I plan to look at it this week. Do you have the urls for those servers too?

helptiger:
> I plan to look at it this week. Do you have the urls for those servers too?

I added the URLs in my last comment

FrankRay78:
## Investigation: why these servers report 0 bps, and why nothing tells you why

Thanks for the detailed report — the specific server URLs and the "works on speedtest.net" contrast were exactly what was needed to trace this. Here's what's happening in the current code.

### The headline

[...]

PR #237, opened 2026-08-28
Profile XmlDoc File Sharing

FrankRay78:
## Why

`ProfileXmlDocTests` failed intermittently, and only ever on a solution-level `dotnet test ./src`. The failure was never captured in a TRX, so the starting point was a hypothesis rather than a finding.

It reproduces deterministically: `XDocument.Load(string)` opens the file with `FileShare.Read` — a share mode that refuses to coexist with an existing writer — so the read throws `IOException: The process cannot access the file because it is being used by another process` whenever another process holds `NetPace.Core.xml` open for writing. That file is a build-copied artefact sitting in the test project's own output directory, so an overlapping build can hold it.

Verified before changing anything: holding a competing write handle reproduced the exact exception, and reverting the fix still fails the new test today.

## What changes

Test-only. No production code changed.

- The XML doc is opened with `FileShare.ReadWrite` instead of via `XDocument.Load(string)`, so a concurrent writer can no longer fail the read.
- A regression test holds a competing write handle and asserts the load still succeeds — a deterministic assertion in place of a timing-dependent race.

**The regression test only discriminates on Windows.** Only Windows enforces share modes; on Unix .NET does not, so the pre-fix load would have succeeded there. The test still runs and still asserts the outcome on every platform — nothing is skipped (Principle X) — but since CI is `ubuntu-latest` only, **CI would not catch a revert of `LoadXmlDoc`**. The premise is asserted under `OperatingSystem.IsWindows()` rather than assumed, so the test reports if it ever stops proving anything on the platform where it does discriminate. This is documented at the site. Adding a Windows CI leg would close the gap but changes CI cost for the whole repo, so it is left as a deliberate follow-up rather than folded into this branch.

**The test uses a private temp file, not the real build artefact.** An earlier draft opened `NetPace.Core.xml` itself for writing, which risked the test contending for the very file the suite reads. The file-sharing behaviour under test is identical on any path.

**This pins a mechanism deliberately** (Principle IX regression exception) — it exists to stop a specific past failure returning.

## How to verify

- [ ] `dotnet build ./src` — zero warnings
- [ ] `dotnet test ./src` — 617 tests, 0 skipped
- [ ] On Windows, revert `LoadXmlDoc` to `XDocument.Load(xmlPath)` and confirm `LoadXmlDoc_WhileAConcurrentWriterHoldsTheFile_StillLoadsTheDocument` fails with `IOException`
- [ ] Confirm no production code is touched

## Related

Pre-existing, introduced in #199 Profile CLI Switch. Unrelated to #206.

Separately observed while soak-testing this branch: `OoklaSpeedtestTests.GetServerLatencyAsync_..._MultipleTestIterations` is independently flaky — it asserts a wall-clock latency within ±25% of a simulated delay and overshoots under machine load (142ms and 85ms against a 45–75ms band, twice in an 8-run soak). Not touched here; worth its own issue.

### DIFF
--- a/src/NetPace.Core.Tests/ProfileXmlDocTests.cs
+++ b/src/NetPace.Core.Tests/ProfileXmlDocTests.cs
@@ -1,4 +1,3 @@
-using System.Xml;
 using System.Xml.Linq;
 using Shouldly;
 
@@ -7,13 +6,10 @@ namespace NetPace.Core.Tests;
 /// <summary>
 /// Verifies <see cref="Profile.Mega"/>'s XML documentation includes the undocumented-payload
 /// caveat, so the warning ships to NuGet consumers via NetPace.Core.xml, and that the loader
-/// used to read that file tolerates a build concurrently rewriting it.
+/// used to read that file tolerates a concurrent writer.
 /// </summary>
 public sealed class ProfileXmlDocTests
 {
-    private const int LoadAttempts = 5;
-    private const int LoadRetryDelayMs = 50;
-
     [Fact]
     public void Profile_Mega_XmlDoc_DocumentsBonusPayloadDependency()
     {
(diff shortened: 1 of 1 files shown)


FrankRay78:
- Bugs or correctness issues

FrankRay78:
@claude How come your review is failing?

FrankRay78:

- Bugs or correctness issues

PR #240, opened 2026-08-29
Renormalise .cs line endings to LF

FrankRay78:
## What

Runs `git add --renormalize` over `*.cs` so the index matches the `*.cs text eol=lf` rule.

## Why

PR #236 added the rule to `.gitattributes` but did not renormalise the files already stored with CRLF. The rule therefore had no effect on the 18 existing CRLF files — and because the attribute now disagrees with what is stored, **git reports all 18 as modified in a clean checkout**. Anyone merging main right now inherits a permanently dirty working tree.

It also leaves a live trap: a scripted edit that rewrites one of those files silently flips its endings and turns a three-line change into a whole-file diff. That happened on #222 — a 3-line edit to `OoklaSpeedtestTests.cs` produced a 3,135-line diff.

## This is line endings only

```
git diff --ignore-cr-at-eol    # empty
```

`git ls-files --eol` now reports `i/lf` for all 84 tracked `.cs` files. Build clean, 0 warnings, 617 tests pass.

**Reviewers: enable "Hide whitespace" on the diff — there is nothing else to see.**

## Deliberately out of scope

`*.sln`, `*.csproj`, `*.props` and `launchSettings.json` are still stored CRLF with no `.gitattributes` rule governing them, so their endings depend on each developer's local `core.autocrlf`. Same trap, narrower blast radius. Worth a follow-up.


FrankRay78:
- Bugs or correctness issues

Issue #239, opened 2026-08-28
Server selection rejects reachable servers and only ever compares one candidate


FrankRay78:
## Summary

NetPace frequently fails to choose a speed test server, reporting "No servers available" on a connection where servers are plainly reachable — and when it does succeed, it does not actually compare the candidates it discovered. The cause is that the check run against each candidate is as expensive as a full latency measurement, so no sensible time budget can be set for it. Screening candidates and measuring the winner should become two different jobs.

## Motivation

Observed on a working connection, minutes apart:

```
> netpace servers -l
  ... 10 London servers listed, latencies 24ms to 106ms, five showing "-"

> netpace --verbosity Debug --profile Small
No speed test servers were found.
```

The table and the run disagree because they probe differently, and because the budget applied during selection is smaller than the probe it is applied to.

A latency probe is all-or-nothing: it performs a fixed number of requests spaced by a fixed interval, returns no partial result, and throws if cancelled part-way (`src/NetPace.Core/Clients/Ookla/OoklaSpeedtest.cs` — `GetServerLatencyAsync`). With the current defaults in `src/NetPace.Core/Clients/Ookla/Settings/LatencyTestSettings.cs` that is roughly a second of deliberate waiting before any network time is counted.

Selection then applies a shrinking budget to that probe (`GetFastestServerByLatencyAsync`): the first candidate gets the full discovery timeout from `src/NetPace.Core/Clients/Ookla/Settings/ServerDiscoverySettings.cs`, and every candidate after the first is given a fraction of the best latency seen so far, subject to a floor that is far below the probe's own minimum runtime.

Three consequences follow:

- Only candidates well under a fraction of the first budget can ever pass. Servers the table shows as reachable at 66ms and 106ms cannot qualify.
- Every candidate after the first is judged against a budget shorter than the probe's own floor, so it always fails. "Fastest server by latency" therefore compares exactly one candidate — whichever responded first — and the ranking is an illusion.
- When no candidate clears the first budget, selection throws and the whole run ends with "No servers available". This needs only a modestly slower link to happen, so it is a routine outcome rather than an edge case.

Selection also probes candidates one after another, and nothing caps how many servers discovery returns, so the worst case scales with the size of the server list.

[...]

PR #241, opened 2026-08-29
Add /build — issue to green committed branch, ahead of /ship

FrankRay78:
## Why

`/ship` starts from "implementation looks done". Nothing in the repo covers the stage before it: point something at a refined GitHub issue and have it come back with a green, committed branch, unattended. `/build` is that stage. You give it an issue number and nothing else; you run `/ship` yourself afterwards, and authorise the merge yourself.

## Why not `/speckit-implement`

It is a different stage of a different pipeline, and four things stop it being adapted:

- **It cannot take an issue as input.** `check-prerequisites.sh --require-tasks` hard-exits without `specs/NNN/{plan,tasks}.md`, so using it means running `specify → plan → tasks` first — by which point every decision `/build` exists to make unattended has already been made by hand.
- **It stops and asks.** Incomplete checklists halt it with a yes/no prompt.
- **It cannot be tailored.** `settings.json` denies `Edit(.claude/skills/speckit-*/SKILL.md)`, and a `--force` re-init resets those files anyway. Nothing NetPace-specific — RED evidence, XML docs on public `Core` APIs, the README `--help` snapshot, AOT-safety, the no-new-`Core`-dependency rule — can ever live in it.
- **Its TDD is ordering, not RED.** "Execute test tasks before their corresponding implementation tasks" never runs a test to watch it fail, which is what Principle I requires.

`/build` therefore sits alongside the spec-kit pipeline rather than wrapping it: it creates no `specs/` folder, so the traceability Stop hook stays a clean no-op and `/raise-pr`'s spec cleanup finds nothing to delete.

## Criteria and labels are independent

The command reads two properties of the issue body separately. An issue may carry either, both, or neither — they are not alternative modes.

- **What to implement.** An `## Acceptance criteria` checklist, when present, is the criteria — that is the shape `/speckit.draftissue`'s template mandates. Otherwise the criteria are derived from the body and written into the final report so the reading can be checked. Where a `## Capability` section's scenarios are also present, they are context for the shape of the change; the checklist is the fuller list and the one to satisfy.
- **How to label the tests.** `**Scenario: X**` labels, where an author used them, get matching `// SCENARIO: X` markers, preserving the Principle VIII chain with the spec/test-plan hop collapsed out. No labels means no markers: an invented label looks like a traceability key and traces to nothing.

Provenance is the wrong axis to split on. #239 has been fully through `draftissue` + `confirmissue`, has no spec folder, and carries three Scenario labels *and* a twelve-item checklist — so "non-speckit issues only" would have excluded the best-refined issue in the repo, and an either/or between the two sections would have dropped nine of its twelve criteria.

## Deliberate omissions

- **No size or complexity check.** Choosing between a direct build and the spec-kit chain belongs to whoever drafts the issue and invokes the command; that call has already been made twice by the time `/build` runs. A pre-flight guess would refuse good issues and wave through bad ones. Instead the command builds what it is handed and reports the wall it hits — "the criteria do not determine the design" is a legitimate `FAILED reason`, discovered by working rather than predicted.
- **No `dotnet format`.** That is `/ship` step 1a. Running it here would leave a format commit that `/ship` step 3 cannot attribute, breaking its "anything in the tree is a review edit" invariant.
- **No push, PR, merge, or `/ship`.** It leaves a clean tree and a committed branch, which is exactly `/ship`'s step 0 precondition.

Two carve-outs from the unattended contract, because `CLAUDE.md` demands discussion for both: a public `NetPace.Core` API change proceeds only if the acceptance criteria require it (and is called out prominently in the report), and a new `NetPace.Core` dependency stops the run outright.

## How to verify

- [ ] `/build` with no argument asks for an issue number and waits; `/build 239` starts straight away.
- [ ] On a labelled issue, the tests it writes carry `// SCENARIO:` markers matching the issue's labels exactly.
- [ ] On an unlabelled issue (#206), no markers appear and the derived criteria are in the final report.
- [ ] The RED step is evidenced — the report quotes real failing-test output from before any production code.
- [ ] It leaves a clean tree on a `feature/*` branch, and `/ship` runs from there without complaint.
- [ ] No `specs/` folder is created, and the traceability Stop hook stays silent.


FrankRay78:
- Bugs or correctness issues

FrankRay78:
Thanks — the finding was productive, though not in the way it was framed. Taking it in three parts.

**The central claim doesn't hold.** #239 does carry `**Scenario: X**` labels — three of them, at lines 46, 49 and 52, under `## Capability`:

```
**Scenario: A run auto-selects a server**
**Scenario: A user lists available servers**
**Scenario: No server can be reached**
```

The review flagged this gap itself — *"Worth confirming directly against issue #239's actual body before merging"* — and then reasoned from the command templates instead. The issue body is ground truth here, because `/build`'s classification reads issue bodies, not templates. So "the traceable branch may rarely (or never) trigger" doesn't follow, and the second How-to-verify item is runnable exactly as written.

**The sub-claim about the templates is correct.** `speckit.draftissue.md` mandates a checklist under `## Acceptance criteria` and "1–3 scenarios" under `## Capability`, without prescribing the bold-label format; `confirmissue.md` never emits it. So "the shape `/speckit.draftissue` produces" over-claimed, and that wording is gone.

**Chasing it down found a real defect the review missed.** #239 carries three Scenario labels **and** a twelve-item `## Acceptance criteria` checklist. Step 3 said the labels *were* the acceptance criteria, so `/build` would have implemented three flows and silently dropped nine criteria — among them `--server <url>` staying unaffected, the USER_GUIDE.md update, the README `--help` regeneration, and the `docs/architecture/` note that exists specifically so a later change cannot remove the selection ceiling.

The two were never alternatives. `acfe9b1` makes them independent properties, either, both or neither of which an issue may carry: the `## Acceptance criteria` checklist says what to implement, and Scenario labels — where an author used them — say what to mark the tests with. The PR body's corresponding section has been rewritten to match.

@claude Please re-review this pull request at `acfe9b1`. Focus on whether Step 3 now describes the two properties accurately against the real issue bodies in this repo (check #239 and #206 directly rather than the command templates), and whether anything else in `build.md` makes a factual claim about the repo's tooling that does not hold.

PR #244, opened 2026-08-31
Capture Learnings Lenses

FrankRay78:
## Why

Running `/capture-learnings` at the end of the #222 review session produced five memory entries and missed the two most reusable things in the whole session — a standing project fact stated in passing ("pre-V1, I don't care about NuGet breaking changes yet"), and two occasions where a rule already written in `CLAUDE.md` produced the wrong answer and I followed it anyway.

Neither was a correction, and the command scanned only for corrections. Three of its four signal classes were phrased negatively, which made it a regression-avoidance tool wearing a knowledge-capture label.

## What changes

**Two new conversation lenses.** *Questions as defects* — a "why is this X?" is an artefact that failed to explain itself, so the fix belongs in a name or a comment, not in Claude's memory. *Reversals* — a position stated then abandoned, separating "I was wrong" from "I caved without new evidence", which look identical in a transcript. Standing project facts now count explicitly as constraint reveals: a versioning posture is stated once and costs every later session if missed.

**A mechanical git lens that needs no transcript.** Churn (a file rewritten 3+ times names a question that kept being re-answered; sibling files that always move together name an unfactored shape) and discarded work (revert pairs are decisions made and unmade). On #222 this immediately found that "where does the all-failed notice live?" had been re-answered four times across `SpeedTestCommand` and `DefaultConsoleWriter` — invisible to a correction scan, because nobody complained.

**An interpretation lens.** *Rule collision* — did an existing written rule produce the wrong answer? *Who caught it* — a defect a human found by eye is a hole in the automated net, and the case worth hunting is a defect the suite had **encoded** rather than caught (a wrong value baked into a snapshot passes forever).

**The output side, which was the real gap.** Every destination previously ended in "add a memory" or "add a mechanism", so a finding of *this documented rule is wrong* had nowhere to land. Amending the rule is now the first-choice destination, ahead of a mechanism, ahead of memory.

Also: the candidate cap scales with session size instead of a flat 5, and step 8 adds an invalidation sweep so a session can retire a memory it made stale.

## Non-obvious

- The valuable half of this diff is the destination ordering, not the lenses. Countering a miscalibrated rule with a memory entry leaves two conflicting instructions in context; amending the rule removes one.
- The lenses were derived from the #222 session and then re-run against it, so their firing there demonstrates wiring, not generality. Worth a sceptical eye on the next unrelated branch.
- Constitution amendments arising from the new "amend rule" path carry a version-bump obligation — step 9 points at the Governance section rather than restating it.

## How to verify

- [ ] Run `/capture-learnings` on a branch with several commits and confirm Lens B runs without a transcript.
- [ ] Confirm the churn one-liner works in this repo: `git log main..HEAD --name-only --pretty=format: -- src docs | grep -v "^$" | sort | uniq -c | sort -rn | head -12`
- [ ] Check the candidate output now carries `Lens:` and `Destination:` fields, and that at least one candidate can route to "Amend rule".
- [ ] Sanity-check that step 7 still blocks on user confirmation before anything is written.

## Related

Gaps surfaced during the review of #222.


FrankRay78:
- Bugs or correctness issues

FrankRay78:
@claude Please address the two review findings above, make only the minimal changes required.

FrankRay78:
@claude Please review the fixes made above, following the earlier review findings.

PR #245, opened 2026-09-01
Upload Redirect 307

### LINKED ISSUE #206: some servers show an upload speed of 0 bps, unlike the speetest.net website

FrankRay78:
## Why

Ookla is migrating its speed test fleet to HTTPS. A migrated server answers a plain-HTTP upload POST with a `307` to its HTTPS endpoint — and it sends that redirect after roughly 64 KB, closing the request stream while the body is still being written.

Since **0.14.0** (#69) the upload body has been streamed in 8 KB chunks instead of handed over as one buffered `ByteArrayContent`. A streaming body cannot survive that mid-write teardown: it fails with `IOException: Unable to write data to the transport connection`, so .NET never reads the `307` and never follows it. Every upload request failed and the run reported `0 bps`.

Roughly ten months of releases (0.14.0 → 0.24.0) have reported `0 bps` against any migrated server — silently, until #222 started surfacing the failure count.

Downloads were never affected, which is what made this look server-specific rather than transport-shaped: a GET carries no body, so it replays across the redirect for free. The same server would report 291 Mbps down and 0 bps up.

## What changes

- The upload endpoint is resolved once, before the measured uploads begin, and uploads go straight to wherever the server actually wants them. Redirects are followed explicitly (up to 5 hops) or honoured by the handler, whichever happens first.
- Upload throughput is now reported correctly against HTTPS-migrated servers.

- **The probe deliberately sends no body.** A redirect is decided by scheme and host, not payload, so an empty POST draws the same answer with nothing to write and nothing to tear down mid-write. It also keeps the probe out of the measurement — an earlier 1 KB probe was caught by `GetUploadSpeedAsync_ShouldReturnSpeedTestResult_WhenSuccessful`, which tallies every byte the server receives against reported throughput. That test passes unmodified here, and is worth keeping that way.
- **#69's memory fix is retained.** Per-chunk streaming is untouched; this only changes *where* the chunks are sent. Buffering the body again would have fixed the bug but re-opened the RAM growth #69 closed.
- **Probing is best-effort.** If it cannot complete, the configured URL is used and any real fault surfaces through the upload requests as before — a broken probe never fails a test run on its own.
- **The regression test pins the outcome, not the mechanism** (Constitution IX): a redirecting upload endpoint must still yield throughput. It holds for any of the redirect-handling approaches we considered.
- **The hermetic seam has a known limit.** `MockHttpMessageHandler` replaces the transport, so no test can reproduce the actual socket teardown mid-body-write. The test covers the redirect contract; the teardown itself was verified manually against a live migrated server.

## How to verify

- [ ] `dotnet build src` — clean, zero warnings
- [ ] `dotnet test src` — 647 pass, 0 skipped
- [ ] New test fails without the fix: stash the `OoklaSpeedtest.cs` change and run `dotnet test src --filter "FullyQualifiedName~WhenServerRedirectsUploadEndpoint"`
- [ ] Against a live migrated server, upload reports real throughput rather than `0 bps`:
      `netpace --no-download --no-latency --server "http://speedtest-bracknell.vodafone.co.uk:8080/speedtest/upload.php"`
      (measured 176.44 Mbps here, against 0 bps / 30 of 30 failed before)
- [ ] Against a non-migrated server (e.g. Aquiss), upload is unchanged

## Related

Regression introduced by #69. Failure counts that made it visible came from #222.

Three follow-ups found while diagnosing this, deliberately left out of this branch:

1. Per-request failure reasons are discarded — only the count survives, which is why this took four releases and a dozen screenshots to pin down. Surfacing them means adding failure detail to `SpeedTestResult`, a public `NetPace.Core` API change.
2. Downloads still pay a redirect per request against migrated servers. Not broken, but silent overhead on every GET.
3. `--no-latency` selects `servers.First()` unvalidated. It is what made this bug reproducible, and the server it lands on shifts between releases.

### DIFF
--- a/src/NetPace.Core.Tests/OoklaSpeedtestTests.Redirects.cs
+++ b/src/NetPace.Core.Tests/OoklaSpeedtestTests.Redirects.cs
@@ -1,0 +1,65 @@
+using System;
+using System.Net;
+using System.Net.Http;
+using NetPace.Core.Clients.Ookla;
+using RichardSzalay.MockHttp;
+using Shouldly;
+
+namespace NetPace.Core.Tests;
+
+/// <summary>
+/// Upload behaviour when a speed test server redirects the upload endpoint.
+/// <para>
+/// Regression cover: Ookla is migrating its fleet to HTTPS, and a migrated server answers a
+/// plain-HTTP upload POST with a 307 to its HTTPS endpoint. From 0.14.0 (PR #69, which replaced
+/// buffered <c>ByteArrayContent</c> with a streaming body) every upload against such a server
+/// failed, reporting 0 bps. The observable contract pinned here is throughput, not the mechanism
+/// by which the redirect is honoured.
+/// </para>
+/// </summary>
+public sealed partial class OoklaSpeedtestTests
+{
+    [Fact]
+    public async Task GetUploadSpeedAsync_ShouldMeasureThroughput_WhenServerRedirectsUploadEndpoint()
+    {
+        // SCENARIO: Upload endpoint answers with a redirect
+
+        const string uploadUrl = "http://example.com/speedtest/upload.php";
+        const string redirectedUrl = "https://migrated.example.com/speedtest/upload.php";
+
+        // Given a server that redirects every upload POST to its migrated endpoint,
+        // and a migrated endpoint that accepts the upload.
+        using var mockHttp = new MockHttpMessageHandler();
+
+        mockHttp.When(HttpMethod.Post, uploadUrl)
+            .Respond(_ => new HttpResponseMessage(HttpStatusCode.TemporaryRedirect)
+            {
+                Headers = { Location = new Uri(redirectedUrl) }
+            });
+
+        mockHttp.When(HttpMethod.Post, redirectedUrl)
+            .Respond(_ => new HttpResponseMessage(HttpStatusCode.OK));
+
+        var httpClient = mockHttp.ToHttpClient();
+        var settings = new OoklaSpeedtestSettings
+        {
+            UploadTest = new()
+            {
+                UploadIncrements = 1,
+                UploadSizeIterations = 4,
+                UploadParallelTasks = 2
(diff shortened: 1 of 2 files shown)


FrankRay78:
- Bugs or correctness issues

FrankRay78:
@claude Review the newly pushed commit d0a2b30, which addresses your earlier review. Please check:
- `ResolveUploadUrlAsync` relocated out of `#region Static Functions`
- New test `GetUploadSpeedAsync_ShouldMeasureThroughputAtTheConfiguredUrl_WhenResolvingTheEndpointFails` covering the best-effort fallback — is the seam right, and does it actually pin the contract?
- The new `IsRedirect` remarks explaining why 301/302/303 are treated alike

Two of your suggestions were deliberately not actioned, and I'd welcome a challenge if you disagree:
- **No test for `maximumHops`** — it is a counted `for` loop, so a test asserting it terminates would be testing the language rather than NetPace.
- **Broad `catch (Exception)` left as-is** — a best-effort probe must not fail the run for any reason, matching the existing justified pattern in `GenericTestSpeedAsync`.

Note: the full suite was run locally — 648 pass, 0 fail, 0 skipped, 0 warnings. The fallback test was also confirmed to fail when the fallback is removed.

Issue #235, opened 2026-08-27
Harness sync: four portable items (piped gate exit codes, tooling install/report, decisiveness memory, transcript retention)


FrankRay78:
## What this is

Four small harness items proven in other repos sharing this harness, each re-grounded in NetPace's own current state. They are independent of one another and unrelated except in size — one PR can carry all four, or item 2 can be split out. None touches `src/`.

Ordered by value.

---

## 1. A piped test command silently fakes a green gate

**Add** `.claude/memory/feedback_background_script_output.md`.

A shell pipeline reports the *last* command's status. So this:

```bash
dotnet build ./src && dotnet test ./src 2>&1 | tail -35
```

exits 0 whenever `tail` succeeds — **including when the suite failed**. This has happened in practice on this harness: a run reported "exit code 0" over a real failure (1 failed / 517 passed), caught only by reading the output rather than trusting the status. A test gate whose exit code is `tail`'s is not a gate.

Why it matters here specifically. `/ship` step 1 says *"Gate on the run's **exit code**, not on any stored marker."* That instruction is correct and it **is** the gate — and a single `| tail` in how the suite gets invoked silently voids it. To be clear about the current state: NetPace is not broken today. All three hooks (`green-gate.sh`, `no-skipped-tests.sh`, `traceability-gate.sh`) set `-uo pipefail`, and `/ship` as written does not pipe. **This item is prophylactic.** The exposure is an agent hand-running a piped suite command inside a session, which is exactly how it went wrong elsewhere — a place no hook is watching.

The same entry covers the second failure mode of the same construct: `tail` buffers its entire input until upstream closes, so a backgrounded long-running script piped through it leaves a 0-byte output file for its whole run. A live PID with an empty log reads as "stuck" when it is merely buffering.

Rules to record:

- Never pipe a test or gate command at all — run it raw and read the output.
- If a pipe is genuinely unavoidable, `set -o pipefail` first, or the exit code belongs to the filter rather than the suite.
- Judge a suite by its `Passed!` / `Failed!` lines, not by exit status alone.
- Background invocations emit raw stdout; apply filtering at the reader, never at the source.

---

## 2. Nothing checks whether the token/context tooling is actually installed

**Add** `scripts/plugin-report.sh` (read-only report) and `.claude/commands/install-harness-tooling.md` (the installer).

[...]

PR #246, opened 2026-09-02
Ship Namespaced Reviewer Agents

FrankRay78:
## Why

`/ship` step 2 named the six `pr-review-toolkit` reviewers bare — `code-reviewer`, `silent-failure-hunter`, and so on. Those ids do not resolve; the plugin registers each as `pr-review-toolkit:<agent>`.

The failure mode is the bad kind. An agent following the instruction either errors out (loud, recoverable) or reads the available-agents list in the error and quietly substitutes `general-purpose` — a review that still runs, still reports, and no longer has any of the specialist prompts behind it. Nothing in the output would say so.

- **Only line 43 was load-bearing.** A repo-wide sweep found the six names in exactly two places: the spawn instruction (line 43) and one prose mention of `code-simplifier` (line 52). The prose one cannot break a spawn, but it is prefixed too so there is a single convention and no ambiguity for the next reader.
- **A sentence was added to line 43** — *"The `pr-review-toolkit:` prefix is required — the bare names do not resolve"* — so anyone later tidying the names back to something shorter has the reason in front of them.
- **This was only reachable after a separate fix.** The plugin had been installed with `--scope project` pinned to an unrelated repository, so it was enabled in settings and loaded nowhere. It is now installed at user scope. That was a change on the box, not in this repo, and is not part of this PR.
- **Three agents self-report a different name than their registered id** — `comment-analyzer` says "code-comment-analyzer", `pr-test-analyzer` says "test-coverage-analyzer", `silent-failure-hunter` says "error-handling-auditor". Harmless, since spawning uses the registered id, but worth knowing if anything ever matches on the self-reported name.

## How to verify

- [ ] Parse the ids straight out of the file rather than retyping them — expect six:

  ```bash
  sed -n '43p' .claude/commands/ship.md | grep -oE 'pr-review-toolkit:[a-z-]+' | sort -u
  ```

- [ ] Spawn each parsed id. All six should resolve and return.
- [ ] Negative control: spawn bare `code-reviewer`. It should fail with `Agent type 'code-reviewer' not found`.
- [ ] Confirm no bare names remain anywhere — expect no hits:

  ```bash
  grep -rnE '\`(code-reviewer|silent-failure-hunter|pr-test-analyzer|type-design-analyzer|comment-analyzer|code-simplifier)\`' --include='*.md' .
  ```

All four were run before raising this. What is **not** covered is a full `/ship` run: this proves the six ids resolve and the agents start, not that step 2 produces useful findings on a real diff. That behaviour is unchanged by this PR.

## Related

Surfaced while reviewing #235, where the underlying finding is recorded under **Confirmed decisions**. The fix itself is deliberately kept out of that issue — #235 is a four-item harness sync and this is a fifth, unrelated defect.

https://claude.ai/code/session_01UmcTTPDWWeAmUfkqgDiNA4

### DIFF
--- a/.claude/commands/ship.md
+++ b/.claude/commands/ship.md
@@ -40,7 +40,7 @@ Read `CLAUDE.md` for project context before proceeding.
    - **Green ⇒ continue.**
 
 2. **Clean-context review (synchronous — this is Review A).** Spawn independent, clean-context reviewer subagents over the branch diff (`git diff main...HEAD`), then run a `/review-slop` pass. The clean-context subagents must not see the code being written — "do not inline the review" governs the *reviewing*. The *deciding-and-fixing* legitimately happens in `/ship`'s own main loop.
-   - Spawn the `pr-review-toolkit` reviewers that apply to this diff. Most **report** findings — `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `type-design-analyzer`, `comment-analyzer` return findings; only `code-simplifier` edits files directly. Launch them in parallel (independent, clean-context Task subagents).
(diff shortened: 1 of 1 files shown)


FrankRay78:
- Bugs or correctness issues

PR #247, opened 2026-09-03
Harness Sync

### LINKED ISSUE #235: Harness sync: four portable items (piped gate exit codes, tooling install/report, decisiveness memory, transcript retention)

FrankRay78:
Four small harness items proven on other repos sharing this harness, each re-grounded in NetPace's own current state. None touches `src/`.

Closes #235

## Why

Each item closes a gap where the harness was trusting something nothing checked:

- A shell pipeline exits with its *last* command's status, so `dotnet test … | tail` reports green over a failed suite. `/ship` step 1 gates on that exit code — a single `| tail` in how the suite gets invoked silently voids the gate, and nothing reports that it has.
- NetPace's config traces `rtk` and `context-mode` right through the harness (allow-entries, `enabledPlugins`, an `rtk` prefix-strip in `green-gate.sh`), but nothing verified either was installed. A tool that silently isn't there costs what one that is there costs; you just stop getting the benefit.
- Transcripts were on the 30-day default. On another box on this harness that default had already deleted a project's entire early-phase sessions — the exact period you want when writing up how the harness performed.

## What changes

- **Memory entry on piped gates** — prophylactic. NetPace is not exposed today (all three hooks set `-uo pipefail`, `/ship` does not pipe); the gap is an agent hand-running a piped suite command mid-session, where no hook is watching.
- **`scripts/plugin-report.sh`** — manually-run, read-only report over TOOLING / CONFIG / HOOKS / PERFORMANCE for `read-once`, `context-mode`, `rtk` and `pr-review-toolkit`. Not a gate: no `--check`, no exit-code contract, wired into no hook, CI job or `/ship` step.
- **`/install-harness-tooling`** — reads each upstream `install.sh` and *prints* the command for a human to run, so the `curl`/`wget` denies stay intact; stops for a settings diff at each point an installer would write a `PreToolUse` hook.
- **Memory entry on decisiveness** — give the verdict the evidence already supports; build within a decision already made. The issue's third rule ("don't over-claim") is folded into the existing `feedback_plain_language_decisions` entry rather than duplicated.
- **`"cleanupPeriodDays": 90`** — one line in `.claude/settings.json`.

- **This branch has no tests, and no RED step happened.** It adds no C# and touches nothing under `src/`; the one testable artefact is the shell script, and #235's confirmed decisions explicitly rejected a `plugin-report.tests.sh` ("a one-off manually-run script, not a gate"). The `/ship` test-coverage reviewer accepted that reading — `scripts/` is outside the `.claude/hooks/*.tests.sh` convention, and `scripts/git-red-phase-commit.sh` is the precedent — but recommended one assertion if any is ever written: **no line of TOOLING/CONFIG/HOOKS output may contain an absolute path or `$HOME`**, which is the diffability contract in one `grep`. Worth considering; deliberately not done here.
- **Review A found the script could not do its one job**, and the fix commit is larger than the original. Eight distinct "I could not look" conditions all collapsed into a confident `no`/`yes`/`ok:` — jq absent, an unparseable settings or registry file, a failed probe. Verdicts are now tri-state (`unknown` never means `no`; `n/a` marks cells with no probe). Each defect was reproduced against synthetic fixtures before and after the fix.
- **`scripts/plugin-report.sh` is committed 644, not 755** like `scripts/git-red-phase-commit.sh`. `Bash(chmod:*)` is on this repo's deny list and I chose not to route around it. Nothing needs the bit — every invocation site uses `bash scripts/plugin-report.sh` — but `chmod +x` it if you want directory consistency.
- **The issue's note that `pr-review-toolkit` is "not loaded in NetPace today" is now stale.** It was installed at project scope for this repo on 2026-09-02, and the report prints `installed=yes enabled=yes`. No deliverable changes: the script reports live state and no doc records install status.
- **`declared` deliberately counts docs, not just config.** Applied to all four rows so the column means one thing throughout — `read-once` is referenced by no config anywhere, only by the guides.

## Known issue, not addressed here

`dotnet format whitespace ./src/NetPace.sln` rewrites **all 79 `.cs` files**, because `.editorconfig` sets `insert_final_newline = false` while every committed file ends with a newline. Reproducible on `main`; unrelated to this branch. `/ship` step 1a would have committed it, but a repo-wide rewrite of every source file's last byte is a second mission and a maintainer decision — so the format churn was reverted and the branch shipped clean. There is genuine drift mixed in (`[ "x" ]` → `["x"]`, trailing whitespace on blank lines), so it is worth its own issue: decide whether `insert_final_newline = false` is intended, then run format once on its own branch.

## How to verify

- [ ] `bash scripts/plugin-report.sh` — four sections, exit 0, nothing written or installed
- [ ] Confirm no absolute path or `$HOME` appears in TOOLING/CONFIG/HOOKS output
- [ ] Break `.claude/settings.json` temporarily — CONFIG must say `unparseable-settings`, not `ok:`, and HOOKS must say why rather than printing an empty section
- [ ] `PATH` without `jq` — TOOLING must read `unknown`, not `no`, with a warning banner
- [ ] `.claude/hooks/*.tests.sh` all pass; `dotnet build ./src && dotnet test ./src` green (648 passed, 0 skipped)

https://claude.ai/code/session_014BzW2ujeWxjL67WZEPZ99i


FrankRay78:
- Bugs or correctness issues

Issue #249, opened 2026-09-03
dotnet format rewrites all 79 .cs files: insert_final_newline disagrees with every stored file


FrankRay78:
## Problem

`dotnet format whitespace ./src/NetPace.sln` rewrites **every `.cs` file in the solution** — 79 of them — on a clean checkout of `main`. `.editorconfig` line 8 sets `insert_final_newline = false` under `[*]`, while essentially every committed file *ends* with a newline, so the whitespace pass dutifully strips them all.

This makes `/ship` step 1a a trap. Step 1a says to commit whatever formatting produced, before the suite; followed literally on any branch, it folds a repo-wide rewrite of every source file's last byte into an unrelated PR, burying the actual change. On #247 — a four-file harness change touching no C# at all — the format pass produced a 79-file diff, which was reverted rather than committed.

## Measured

On `main`, running the full `dotnet format style` + `whitespace` pair:

| `insert_final_newline` | `.cs` files rewritten |
|---|---|
| `false` (current) | **79** |
| `true` | **23** (25 including the `style` pass) |

The residual 23 are genuine, wanted fixes rather than churn:

- 11 files that truly lack a trailing newline, which format then *adds*
- 6 blank lines carrying trailing whitespace
- collection-expression spacing, e.g. `RunAsync([ "--verbosity", "Minimal" ])` → `RunAsync(["--verbosity", "Minimal"])`

And it converges: a second consecutive `style` + `whitespace` pass changes **nothing**. So one cleanup commit makes `/ship` step 1a a genuine no-op instead of a landmine.

Note the cleanup reaches `examples/ConsoleApp/Program.cs` as well as `src/` — it is one of the files with no trailing newline.

## Recommendation

1. Set `insert_final_newline = true` in `.editorconfig`, so the rule agrees with what is stored and with the POSIX/git convention that makes `\ No newline at end of file` markers stop appearing in diffs.
2. Run `dotnet format style` + `whitespace` once on its own branch and commit the ~25-file result, so the drift lands as a reviewable formatting-only change rather than inside someone's feature PR.

Flipping the setting the other way — accepting `false` and committing the 79-file strip — is the alternative, but it leaves every file without a trailing newline, which noisier diffs and a good deal of tooling will keep complaining about.

## This is the third instance of one class

The same shape has now bitten three times: **a formatting rule that asserts something the stored files contradict.**

- **#236** added `*.cs text eol=lf` to `.gitattributes`.
[...]

PR #253, opened 2026-09-03
Plugin Report Exec And Stats

### LINKED ISSUE #252: plugin-report.sh: not executable, and the read-once PERFORMANCE row runs the wrong subcommand
Two independent defects in `scripts/plugin-report.sh`, both found while running `/install-harness-tooling` on a box where none of the three declared tools were installed yet.

## 1. The script is not executable

`scripts/plugin-report.sh` is committed `100644`, while every sibling in `scripts/` and `.claude/hooks/` is `100755`. Running it the obvious way fails:

```
$ ./scripts/plugin-report.sh
/bin/bash: line 1: ./scripts/plugin-report.sh: Permission denied

$ git ls-files -s scripts/plugin-report.sh
100644 0749346c83e335b13d46507df4671a43f03cc6a7 0	scripts/plugin-report.sh
```

`bash scripts/plugin-report.sh` works, and that is the form `/install-harness-tooling` prints, so the missing bit stays invisible until someone runs the script directly — which is the natural thing to do for a report script sitting next to executable siblings.

Note for sequencing: `git update-index --chmod=+x scripts/plugin-report.sh` sets the mode in the index without invoking the `chmod` binary, so this fix is not blocked by the current `Bash(chmod:*)` deny rule and does not depend on #250 landing first.

## 2. The read-once PERFORMANCE row reports diagnostics, not counters

The PERFORMANCE section describes itself as "Live counters". The rtk row asks for counters; the read-once row asks for the installation diagnostic:

```sh
if [ $rtk_inst -eq "$YES" ]; then perf rtk rtk gain; else echo "rtk: n/a — not installed"; fi
if [ -x "$ro_dir/read-once" ]; then perf read-once "$ro_dir/read-once" verify; else echo "read-once: n/a — not installed"; fi
```

`verify` is a 13-point installation check with no counters in it, and it is already invoked on line 155 as the reachability probe. The result is that PERFORMANCE prints twenty lines of `[ok]` where a savings figure belongs — and, because `verify` exits 0 on a healthy install, it reads as a successful counter report rather than a missing one.

The counters subcommand is `stats` — the one the read-once installer points at on completion. On a box with no captured reads yet it prints `No read-once data yet. Stats appear after your first Claude Code session with the hook installed.`, which is the honest answer the section should be showing.

Line 155 should keep `verify`; only the PERFORMANCE row changes.

## Acceptance criteria

- [ ] `scripts/plugin-report.sh` is `100755`, matching its siblings in `scripts/` and `.claude/hooks/`.
- [ ] The PERFORMANCE section reports read-once's usage counters, or read-once's own "no data yet" message when nothing has been captured — not the installation checklist.
- [ ] The TOOLING section's `reachable=` value for read-once still reflects whether read-once passes its own installation check.

FrankRay78:
Closes #252

## Why

Both defects were invisible in normal use, which is why they survived. The missing exec bit stayed hidden because every documented invocation is `bash scripts/plugin-report.sh`; the wrong subcommand stayed hidden because `read-once verify` **exits 0**, so a counter row full of `[ok]` checklist lines read as a successful report rather than a missing figure. A script whose whole product is a truthful yes/no had committed the exact sin it exists to catch in other tooling.

## What changes

- The PERFORMANCE section reports read-once's counters (`stats`) instead of its diagnostic checklist (`verify`). On a box with nothing captured it now prints read-once's own honest line: `No read-once data yet. Stats appear after your first Claude Code session with the hook installed.`
- `scripts/plugin-report.sh` is now `100755`, matching every sibling in `scripts/` and `.claude/hooks/`. **This is not visible in the content diff** — use `git diff main...HEAD --summary` to see the `100644 => 100755` mode change.

**The two calls to read-once are deliberately inconsistent, and "harmonising" them would reinstate the bug.** Line 155 keeps `verify` (it drives the TOOLING `reachable=` column); line 274 uses `stats`. Issue #252 states the constraint — "Line 155 should keep `verify`; only the PERFORMANCE row changes" — but that constraint existed nowhere in the code, so the review added it to the comment. Note also that `read-once gain` is an alias for `stats`, and the line immediately above is `perf rtk rtk gain`, which makes the wrong refactor look especially reasonable.

**This PR ships no committed tests, deliberately.** Issue #235's confirmed decisions record the author's explicit call: *"plugin-report.sh ... with no plugin-report.tests.sh ... it is a one-off manually-run script, not a gate."* RED-GREEN was driven by a 14-case harness in the gitignored `.claude/scratch/`, which fails 6/12 against `main` and passes on this branch. **That decision is worth a fresh look** — #252 is two defects a committed matrix would have caught, and the harness that caught them is about to be deleted with the scratch tree. Landing that file as-is at `scripts/plugin-report.tests.sh` would be minutes of work, but it has a documented author decision against it, so it is your call rather than something this PR should override.

## How to verify

- [ ] `./scripts/plugin-report.sh` runs directly without `bash` and exits 0.
- [ ] `git diff main...HEAD --summary` shows `mode change 100644 => 100755`.
- [ ] The PERFORMANCE section's read-once row shows counters or a "no data" message, not a list of `[ok]` lines.
- [ ] The TOOLING row still reads `read-once: ... reachable=yes` — that column must still come from `verify`.

## Deferred review findings

Clean-context review raised these. All were confirmed real but concern **pre-existing** code this branch does not touch (the diff is a single hunk, `@@ -271 +271,4 @@`), so they are recorded here rather than folded in per CLAUDE.md's "don't fold a second mission into an in-flight branch".

| # | Finding | Notes |
|---|---|---|
| 1 | `probe()` feeds a **raw exit code** into `yn()`, which maps `3` → `n/a` ("no probe for this cell") and `2` → `unknown`. A hard failure can report as "we never asked". | Safe only by luck today: `read-once verify` uses 0/1 and `rtk --version` returns 0. `rtk` is clap-based and clap uses exit 2 for usage errors. Lines 74-86, 155, 170. |
| 2 | `installed` uses `[ -d ]` (:153) but the PERFORMANCE guard uses `[ -x ]` (:274), so a present-but-non-executable binary prints `installed=yes` **and** `n/a — not installed` in the same run. | Two sections contradicting each other. Thematically apt given this PR. |
| 3 | `NETPACE_CLAUDE_HOME` selects which read-once **binary** runs, but read-once hardcodes `CACHE_DIR="${HOME}/.claude/read-once"`, so the counters always come from the default home. | Newly material now that the row carries home-dependent data. Only bites under the diagnostic override. |
| 4 | A read-once predating `stats` surfaces as `PROBE FAILED (exit 1)`, blaming the tool for the report's own assumption. | Not silent — the cause is printed — but the headline misattributes. Now covered in the harness. |
| 5 | `timeout` exit 124 reports as `PROBE FAILED`, where the file's own rule says a timeout is "could not look" → `unknown`. | Same class as #1. |
| 6 | `"${TIMEOUT[@]}"` on an empty array under `set -u` aborts on bash < 4.4; macOS ships 3.2 and has no `timeout(1)`, so the array is empty exactly there. | Reported by review; **not verified here** (this box runs bash 5.x). Fix is `"${TIMEOUT[@]+"${TIMEOUT[@]}"}"`. |
| 7 | All three documented call sites still use `bash scripts/plugin-report.sh` (`docs/agentic-workflow-NetPace.md:105`, `.claude/commands/install-harness-tooling.md:13` and `:62`). | Flagged by three of four reviewers. Switching them to `./scripts/plugin-report.sh` would make the documented workflow a regression detector for the exec bit. |

One finding was graded **Blocker** and is deliberately not fixed: that `read-once stats` exits 0 both when there is genuinely no data and when its stats file is unreadable, so the row cannot tell them apart. The facts check out, but this is read-once's own behaviour, and AC2 explicitly asks for "read-once's own 'no data yet' message". The proposed fix would couple this script to read-once's private `--json` schema — the exact coupling its header warns against for `.plugins[k][].installPath`. Worth an upstream issue, not a change here.

## Related

- Closes #252
- Formatting (`/ship` step 1a) was **skipped** — #249 is still open, and `dotnet format whitespace` currently rewrites all 79 `.cs` files. This branch touches no C#, so the step was a no-op for it regardless.

https://claude.ai/code/session_01TFuh2edjRehoFbZh2BnXUW

### DIFF
--- a/scripts/plugin-report.sh
+++ b/scripts/plugin-report.sh
@@ -268,6 +268,9 @@ perf() {
   printf '%s\n' "$out" | sed 's/^/  /'
 }
 if [ $rtk_inst -eq "$YES" ]; then perf rtk rtk gain; else echo "rtk: n/a — not installed"; fi
-if [ -x "$ro_dir/read-once" ]; then perf read-once "$ro_dir/read-once" verify; else echo "read-once: n/a — not installed"; fi
+# `stats`, not `verify` (#252). read-once's counters subcommand is `stats`; `verify` is its 13-point
+# installation check, already run above as the TOOLING reachability probe — printing it here filled
+# a counter row with a checklist that exits 0, so a missing figure read as a healthy report.
+if [ -x "$ro_dir/read-once" ]; then perf read-once "$ro_dir/read-once" stats; else echo "read-once: n/a — not installed"; fi
 if [ $cm_inst -eq "$YES" ]; then echo "context-mode: installed — counters are MCP-only; run ctx_stats in-session"; else echo "context-mode: $(yn $cm_inst) — no shell-visible counters"; fi
 echo "pr-review-toolkit: n/a — exposes no counters"
--- a/scripts/plugin-report.sh
+++ b/scripts/plugin-report.sh
@@ -268,9 +268,11 @@ perf() {
   printf '%s\n' "$out" | sed 's/^/  /'
 }
 if [ $rtk_inst -eq "$YES" ]; then perf rtk rtk gain; else echo "rtk: n/a — not installed"; fi
-# `stats`, not `verify` (#252). read-once's counters subcommand is `stats`; `verify` is its 13-point
-# installation check, already run above as the TOOLING reachability probe — printing it here filled
-# a counter row with a checklist that exits 0, so a missing figure read as a healthy report.
+# `stats`, not `verify` (issue #252). The two calls deliberately use different subcommands, and
+# making them agree is the wrong refactor: the TOOLING reachability probe above runs `verify`,
+# read-once's full diagnostic, while counters live under `stats` (`gain` is an alias for it).
+# `verify` here would fill a counter row with a checklist that exits 0, so the missing figure
+# would read as a healthy report.
 if [ -x "$ro_dir/read-once" ]; then perf read-once "$ro_dir/read-once" stats; else echo "read-once: n/a — not installed"; fi
 if [ $cm_inst -eq "$YES" ]; then echo "context-mode: installed — counters are MCP-only; run ctx_stats in-session"; else echo "context-mode: $(yn $cm_inst) — no shell-visible counters"; fi
 echo "pr-review-toolkit: n/a — exposes no counters"


FrankRay78:
- Bugs or correctness issues

PR #255, opened 2026-09-04
EditorConfig Final Newline

### LINKED ISSUE #249: dotnet format rewrites all 79 .cs files: insert_final_newline disagrees with every stored file

FrankRay78:
Closes #249

## Why

`.editorconfig` declared `insert_final_newline = false` while 77 of the 92 non-empty `.cs` files end with a newline. On a clean checkout of `main`, `dotnet format whitespace` therefore stripped the last byte of every file it governs — which made `/ship`'s format step a trap. On #247, a four-file harness change touching no C# at all produced a 79-file diff.

Same shape as #236 and #240: **a formatting rule that asserts something the stored files contradict.** Third instance, so this also closes the class rather than the instance.

## What changes

- `insert_final_newline` set to `true` — the value the overwhelming majority of files already store, and the one that keeps `\ No newline at end of file` out of diffs.
- Snapshot and vendored-spec-kit files are recorded as explicit exemptions with their reasons, instead of being left to disagree silently.
- `.gitattributes` now governs **every** tracked file (`* text=auto eol=lf` plus explicit binary markers). #240 left `*.sln`, `*.csproj`, `*.props` and `launchSettings.json` ungoverned precisely because an enumeration only covers what someone remembered to add.
- CI fails if `dotnet format` would change anything.

**The commits are ordered so the rule change is readable.** `.editorconfig` was itself stored CRLF, so converting it produced a 420-line diff. That conversion sits in the mechanical commit; the actual rule change (`c6028d6`) is **34 insertions, 2 deletions**. Review that one closely and skim the other.

**Nothing behavioural hides in the 40-file mechanical commit.** Diffed with whitespace, CR and BOM normalised away, only three files differ: using-directive reorders in `Program.cs`, `SpeedTestCommand.cs` and `OoklaSpeedtest.cs`, all from `dotnet_sort_system_directives_first`. Plus one BOM removed from `Verbosity.cs`. Verified independently and by a clean-context reviewer.

**Most files in this PR have a deliberately invisible diff.** 34 of the 42 changed files differ only in bytes that do not render: CR characters, trailing spaces, and final newlines. `NetPace.Benchmarks.csproj` and `NetPace.sln` are whole-file CRLF->LF conversions, so every line shows as changed while reading identically. Use GitHub's **Hide whitespace changes** toggle (`?w=1`), or locally:

```bash
git diff main...HEAD -w --ignore-blank-lines --ignore-cr-at-eol --stat
```

That reduces the PR to the files with real content changes. Committing the conversion is not optional: once `.gitattributes` declares `* text=auto eol=lf`, git treats those 9 files as un-normalised, so a fresh clone would report them modified until this lands — the same cleanup #240 had to do for 18 `.cs` files. Deferring it would push a whole-file conversion into whoever next edits `NetPace.sln`, which is the failure this issue exists to stop. One BOM is removed (`Verbosity.cs`, by `dotnet format` under `charset = utf-8`); the `.csproj` BOMs are preserved.

**Note on blame:** those 9 files now attribute every line to the cleanup commit. A `.git-blame-ignore-revs` entry would restore it; not added here.

**No unit tests, by design.** The changes are a config file, `.gitattributes`, a CI workflow step and a CIR — no C# production code. The guard is `dotnet format --verify-no-changes` in CI, which exits **2** against `main` (WHITESPACE errors in `SpeedScale.cs`, IMPORTS errors in three files) and **0** against this branch, in 26 seconds. That is the RED-GREEN evidence, and it covers every rule in `.editorconfig` — `charset`, `indent_size`, the `dotnet_*`/`csharp_*` style rules, the analyzers — rather than a hand-picked subset.

An earlier revision of this branch grew a bespoke `.editorconfig` reader and repository scanner to assert the same thing. It was deleted: ~600 lines of test infrastructure with its own bugs, covering strictly less than the one-line check that replaced it. The CIR records that as a rejected alternative.

**Follow-up, deliberately out of scope:** `charset = utf-8` is declared while roughly 90 stored files carry a UTF-8 BOM. Same rule-vs-storage class as this issue; `dotnet format` does not act on them today, so it is not live churn, but any editor honouring `.editorconfig` would strip them. Worth its own issue.

## How to verify

- [ ] `dotnet format ./src/NetPace.sln --verify-no-changes` exits 0 on this branch, and non-zero on `main`
- [ ] `git status` is clean immediately after a fresh clone
- [ ] `git ls-files --eol` reports no `i/crlf`
- [ ] `dotnet build ./src && dotnet test ./src` — 648 passed, 0 skipped
- [ ] The `Fail if formatting would change anything` step passes in CI on this PR

## Related

- #249 — the issue
- #247 — where this surfaced; the format churn was reverted and that branch shipped clean
- #240, #236 — the two previous instances of the same class

https://claude.ai/code/session_01HusZvrit2hSGrKwQD72UR3


FrankRay78:
- Bugs or correctness issues

FrankRay78:
@claude A further commit has been pushed (`f5a37aa`) — please review the **latest push** specifically, on top of your review of the branch as a whole.

It amends the project's own rules rather than its code:

- **`.specify/memory/constitution.md`** — Principle I (TDD, NON-NEGOTIABLE) gains a *Configuration, tooling and CI changes* carve-out. Version 1.7.0 → 1.8.0.
- **`CLAUDE.md`** — a paired rule beside the existing TDD one.
- **`.claude/commands/build.md`** — step 5 now decides whether the change is production code before writing any test.
- **`.claude/memory/feedback_guard_outgrows_the_fix.md`** — new project memory entry, plus its `MEMORY.md` pointer.

Context for why: an earlier revision of this branch satisfied Principle I by writing a bespoke `.editorconfig` parser and repository scanner (~600 lines) purely so the RED step had something `dotnet test` could execute — larger than the 34-line fix it guarded. It was deleted and replaced by `dotnet format --verify-no-changes` in CI. The amendment is meant to stop that recurring.

Please focus on:
- Does the Principle I carve-out weaken the TDD guarantee anywhere it shouldn't? Specifically, can "this is configuration, not production code" be stretched to skip tests on a change that genuinely needs them?
- Is the constitution version bump correct (MINOR for materially expanded guidance) per its own Governance section?
- Governance step 4 requires noting which downstream documents were reviewed in lockstep. Are `CLAUDE.md`, `.claude/commands/build.md` and `docs/conventions/csharp-style.md` the right set, and are the first two now consistent with the amended principle?
- Any contradiction introduced between the new carve-out and Principle X (No Skipped Tests) or the Testing Standards section.

PR #258, opened 2026-09-04
Subagent Permission Inheritance

### LINKED ISSUE #251: Reviewer subagents do not inherit bypassPermissions, so /ship step 2 prompts 10-15 times
## Problem

`/ship` is designed to run unattended, so it can be driven back-to-back by a loop. It cannot today: **step 2 prompts for permission 10–15 times**, even with the session in `bypassPermissions`. Observed on #247, during the `pr-review-toolkit` reviewer agents, before the PR was raised.

## Evidence

The main thread and the subagents behaved completely differently in the same session:

| | Main thread | The 5 reviewer agents |
|---|---|---|
| `permissionMode` records | **50**, all `bypassPermissions` | **0 — none recorded at all** |
| Bash calls | 114 | 72 |
| Not matched by `permissions.allow` | ~50 | **30** |
| Permission prompts | **0** | the 10–15 observed |

The main thread ran ~50 commands no allow-rule covers (`python3`, `sed`, `jq`, `diff`, `until`, `git fetch`, …) and prompted for **none** — bypass worked there. The reviewers ran 30 of the same character and prompted. The only variable is the subagent boundary.

30 calls producing 10–15 prompts is consistent, since approving a pattern covers later calls matching it. The two `toolDenialKind: "permission-rule"` records from the same agents carry `userType: "external"`, i.e. a user-facing decision path.

**Inference to confirm first:** subagents appear not to inherit the parent session's permission mode, falling back to the project `defaultMode` (`acceptEdits`), where every unmatched Bash command prompts. The mechanism and the correct setting name have not been verified against Claude Code's documentation — that is task one, not an assumption to build on.

## Why an allowlist cannot fix this

What the reviewers actually needed:

```
 13  compound commands led by a variable assignment
  6  cd
  5  bash            (running `bash scripts/plugin-report.sh` to verify it)
  2  jq
  2  sed
  1  awk
  1  a non-allowlisted git subcommand
```

Adding `Bash(bash:*)`, `Bash(jq:*)`, `Bash(sed:*)`, `Bash(awk:*)`, `Bash(cd:*)` covers 10 of 30. The largest bucket — **13 compound commands beginning with a variable assignment** — can never match a `Bash(prefix:*)` rule. That shape is intrinsic to how these reviewers work: they stand up a synthetic environment to exercise failure paths, e.g.

```
PATH="$SB/fakebin:$PATH" NETPACE_CLAUDE_HOME=$SB/roh bash scripts/plugin-report.sh
```

That technique is what found the real defects on #247 — a report that laundered failed probes into confident answers. Constraining reviewers to prefix-matchable commands would remove the capability that made the review valuable, so widening the allowlist is the wrong lever even where it would work.

## Why it matters beyond the noise

- `/ship`'s own docs say it "is designed to run to completion without prompting, so it can be driven by an automated loop". Step 2 breaks that today.
- The failure is silent in the other direction too: an unattended run with nobody watching leaves the reviewers blocked on their most informative commands, so the review quietly gets weaker rather than visibly stopping.

## Acceptance criteria

- [ ] The mechanism is confirmed against Claude Code's documented behaviour, and the setting (or lack of one) that governs subagent permission mode is identified and written down.
- [ ] A `/ship` run started in `bypassPermissions` completes step 2 with **zero** permission prompts.
- [ ] Reviewer agents can still run compound commands with environment-variable prefixes — verified by an agent standing up a synthetic `PATH`/`NETPACE_CLAUDE_HOME` fixture without prompting.
- [ ] If no inheritance mechanism exists, the limitation is recorded in `docs/agentic-workflow-NetPace.md` alongside whatever workaround is adopted, so the next person does not re-diagnose it.

## Related

- PR #247 — where this was observed and diagnosed.
- The permissions issue (`git push` → allow, `chmod` → ask) — a separate cause found in the same investigation; fixing it does not fix this.

FrankRay78:
## Why

`/ship` step 2 prompted 10–15 times per run even in `bypassPermissions`, so it could not be driven unattended despite its docs promising exactly that.

The issue's diagnosis turned out to be wrong, and confirming that was the first acceptance criterion. Subagents **do** inherit the parent's permission mode — *"if the parent uses `bypassPermissions` or `acceptEdits`, this takes precedence and can't be overridden"* — and `permissionMode` frontmatter is ignored for plugin subagents entirely. The evidence the issue rested on (no `permissionMode` records for the reviewers) was a measurement artifact: subagent turns are not written to the project transcript at all, so "zero records" meant *not recorded*, not *no mode*.

The real cause is that `ask` rules sit **above** every permission mode. Nothing clears them — not `bypassPermissions`, not `--dangerously-skip-permissions`. Reviewer subagents build and tear down synthetic fixtures, so teardown hit `Bash(rm:*)` on nearly every run.

## What changes

- `Bash(rm:*)` and `Bash(rmdir:*)` off `permissions.ask` — the whole cause of the prompting.
- `git pull`, `merge`, `rebase`, `reset`, `tag` off it too. None was covered by an allow rule, so in Manual and `acceptEdits` they prompted as unmatched commands with or without the entry; the entry's only effect was to break through `bypassPermissions`. All five are local and leave the old tip in the reflog.
- `Bash(git push:*)` is all that remains — the one command whose blast radius a disposable box does not contain.
- The mechanism, and the headless-oracle technique below, documented in `docs/agentic-workflow-NetPace.md`, with the decision and its rejected alternatives in a CIR.
- `/build` no longer prescribes a GitHub closing keyword for its commit message (see below).

**The commit that fixed this is already on `main`.** `94005cf` was pushed directly, so this PR carries only the follow-on work. That commit's message began `Fix #251:`, which auto-closed the issue on push, before any review — which is the bug `0a0a4fc` fixes in `.claude/commands/build.md`. Step 6 there prescribed `Fix #<N>`, while the same skill's closing report told the invoker to put `Closes #<N>` in the PR body; those contradict, and the commit wins. It is now `Refs #<N>`.

**The same rules bite headless workers silently.** `claude -p` cannot draw a prompt, so an `ask`-matched call is *denied* instead — the worker loses that capability and carries on degraded rather than stopping. A green unattended run is not evidence its `ask` rules were harmless.

**That asymmetry is what verified this PR.** Since headless denies anything that would prompt interactively, a headless run completing every step proves an interactive run would not prompt — a check with no human in the loop to mask it. It is not CI-gateable (it needs the `claude` binary and an authenticated session), so it stays a manual check documented for when the `ask` list changes.

**Safety that is deliberately untouched:** Claude Code still refuses `rm`/`rmdir` against a critical path (`.git`, `.claude`, dotfiles) in every mode, and no allow rule or `PreToolUse` hook can approve past it. What was given up is a confirmation step inside a git-tracked, rebuildable working tree.

Replacing `Bash(git push:*)` with an `allow` plus a `deny` on `--force` is deliberately left to the separate permissions issue rather than folded in here.

## How to verify

- [ ] `dotnet build ./src && dotnet test ./src` — 648 passed, 0 skipped, 0 warnings.
- [ ] `permissions.ask` in `.claude/settings.json` contains exactly one entry, `Bash(git push:*)`.
- [ ] Headless oracle, expecting the first to run and the second to be refused:
      `claude -p --dangerously-skip-permissions "run: git tag --list | head -1"` → RAN
      `claude -p --dangerously-skip-permissions "run: git push --dry-run ..."` → DENIED
- [ ] A reviewer subagent stands up a `PATH`/`NETPACE_CLAUDE_HOME` fixture, drives `scripts/plugin-report.sh` under it, and tears it down with no prompt.
- [ ] `grep -n "Fix #" .claude/commands/build.md` returns nothing in the step-6 commit guidance.

## Related

- Closes #251.
- The `git push` / `chmod` permission entries are owned by the separate open permissions issue.


FrankRay78:
- Bugs or correctness issues

PR #261, opened 2026-09-04
Permissions Push Allow Chmod Ask

### LINKED ISSUE #250: Permissions: allow git push, and downgrade chmod from deny to ask
Two changes to `permissions` in `.claude/settings.json`, both found while investigating why `/ship` interrupted an unattended run on #247.

## 1. `Bash(git push:*)`: `ask` → `allow`

`/ship` is designed to run to completion without prompting, and `/raise-pr` **step 7 is `git push`**. With `Bash(git push:*)` in `ask`, the pipeline stalls at the second-to-last step — after the suite, five reviewer agents, the fix commits and the PR body have all been done. The most expensive possible place to stop.

On #247 this was masked because the session ran in `bypassPermissions`. It is not masked in the project's own `defaultMode`, which is `acceptEdits`.

## 2. `Bash(chmod:*)`: `deny` → `ask`

`chmod` is the odd entry in the deny list. The others are coherent — `sudo`/`su`/`chown` are privilege escalation, `curl`/`wget`/`ssh`/`scp` are network exfiltration. `chmod` is neither: it changes a local file mode and grants no capability a shell does not already have. Meanwhile it has a measurable cost, twice on one PR:

- **It silently blocked a reviewer.** During `/ship` step 2 on #247, `pr-review-toolkit:silent-failure-hunter` was denied twice while building fake `rtk` and `read-once` executables to test whether `plugin-report.sh` laundered a failing probe into output that looked like counters. Two `toolDenialKind: "permission-rule"` records; the agent had to work around it. That was the exact test that found a real defect.
- **It left an inconsistent file mode.** `scripts/plugin-report.sh` is committed `100644` while every sibling in `scripts/` and `.claude/hooks/` is `100755`, because the deny rule made setting the bit impossible and routing around an explicit deny was not appropriate.

Two properties of `deny` make `ask` the better fit here:

- **`deny` is enforced even in `bypassPermissions`.** Confirmed on #247: 50 `permissionMode: bypassPermissions` records in the session, and the `chmod` denials still fired. So a deny rule is a *silent hard block* mid-agent — no prompt, no approval path, and the agent simply proceeds with less evidence. `ask` at least surfaces the decision.
- **The matcher scans the whole command, not just the prefix.** The blocked command began with `SB=/tmp/…`; `chmod` appeared eight lines later inside a heredoc. Since ~87% of commands in that session were compound, a denied binary anywhere in a command line blocks it.

## Acceptance criteria

- [ ] `Bash(git push:*)` is in `permissions.allow`; a `/ship` run reaches `gh pr create` without a prompt for the push.
- [ ] `Bash(chmod:*)` is in `permissions.ask` rather than `permissions.deny`.
- [ ] `permissions.deny` still covers privilege escalation (`sudo`, `su`, `chown`) and network exfiltration (`curl`, `wget`, `ssh`, `scp`) unchanged.

## Not included here

The `green-gate.sh` hook is wired with no `if` guard, so its `statusMessage` renders on every Bash call — 56 times during `/ship` on #247. That is UI noise rather than a permission problem, and the `if`-less wiring is deliberate (documented in `.claude/hooks/README.md`, so its prefix-stripping still catches `cd repo && dotnet test --no-build`). Removing just the `statusMessage` would fix it. Left out to keep this issue to the permission changes.

## Related

- PR #247 — where both problems surfaced.
- The subagent permission-inheritance issue — the separate cause of the prompts seen during `/ship` step 2.

FrankRay78:
## Why

Three permission rules that between them stopped `/ship` running unattended. Two were named in #250; the third surfaced during the `/ship` run that shipped this branch and was scoped into the issue once it was understood.

`git push` sat on `permissions.ask`, so `/raise-pr` stalled at its push step — the second-to-last action of an otherwise complete pipeline, after the suite, the reviewers, the fix commits and the PR body were all done.

`chmod` was the odd entry in `deny`: the rest of that list is privilege escalation and network exfiltration, but `chmod` changes a local file mode and grants no capability a shell does not already have. `deny` is enforced even in `bypassPermissions`, so it was a silent hard block with no approval path, and it cost a reviewer the fixture it needed on #247.

The six `Read(…)` deny rules made any recursive read of the repository escalate to an approval no permission mode grants. A reviewer running `grep -rn … .` hit it during this branch's own review.

## What changes

- `Bash(git push:*)` → `allow`.
- `Bash(chmod:*)` → `ask` (from `deny`).
- The six `Read(…)` rules removed from `deny` — `.env`, `.env.*`, `secrets.*`, `.ssh/**`, `appsettings.*.json`, `appsettings.json`.
- `permissions.deny` keeps all seven of its Bash entries unchanged.
- Three CIRs, and the moves narrated in the workflow doc's permissions section.

**The `Read` rules are removed on a live-exposure argument, and that argument has a shelf life.** The escalation is glob-scope-based, not existence-based — the repo contains none of those files, yet the prompt fired naming `Read(**/.env)` — so there was no partial fix; every rule was `**/`-anchored and matchable somewhere under the working directory. Removing them means Claude can read such a file *if one ever appears*. A scan of the tree found none, plus no key or certificate files and no content matches for cloud keys, forge tokens, private-key headers or quoted credential assignments. If NetPace ever acquires a secret file, this needs revisiting — a `PreToolUse` hook would restore the cover without reintroducing the escalation, since the escalation keys on settings rules rather than hooks.

**That change is unconfirmed by observation, and the CIR says so.** Establishing a RED needs an un-memoised session; approval had already been granted in the session that made the change, and the headless oracle is blind to this class. The reasoning is sound, the mechanism understood, but it wants a fresh-session check — steps in the CIR.

**This branch also corrects a claim it introduced earlier.** "Headless is a permission oracle" is sound for `allow`/`ask`/`deny` matching but blind to escalations that need an interactive surface: these greps prompt interactively and run clean under `claude -p`. It answers "which rule matched", not "would a human have been asked".

**Force-push is now unguarded, deliberately.** The preceding CIR anticipated pairing the `git push` allow with a `deny` on `--force`. Only the `allow` shipped: rules match on textual prefix, so `Bash(git push --force:*)` catches `--force` and misses `-f`, `git push origin main --force`, and every compound form — the same objection that sank a narrowed `rm` rule a day earlier. A guard a flag reorder defeats is worse than none. A real reduction in cover, not an oversight; the honest fix is a parsing hook and wants its own issue.

**`ask` does not deliver what #250 expects, in the case that motivated it.** The issue argues `ask` beats `deny` because it "at least surfaces the decision". True interactively, false headlessly: a `claude -p` worker has no surface to prompt on, so an `ask` rule denies silently — indistinguishable from the `deny` it replaced. Implemented as written; gap documented.

**The issue's file-mode evidence is stale.** It cites `scripts/plugin-report.sh` at `100644` as damage from the deny rule. It is `100755`, corrected in #253, so no repair was needed.

## How to verify

- [ ] `dotnet build ./src && dotnet test ./src` — 648 passed, 0 skipped, 0 warnings.
- [ ] `permissions.ask` is exactly `["Bash(chmod:*)"]`; `Bash(git push:*)` is in `allow`; `deny` has no `Read(…)` entries and still lists the seven binaries.
- [ ] Headless oracle — a headless worker denies what a rule would stop, and the text distinguishes the kinds:
      `git push --dry-run …` → RAN (was: `requires user confirmation`)
      `chmod +x <file>` → `you haven't granted it yet` (was: `matches deny pattern Bash(chmod:*)`) — proving `ask`, not `deny`
      a denied network binary → `Blocked by security policy` (unchanged)
- [ ] **In a fresh interactive session**, as the first such command of that session: `grep -rn "<any term>" .` from the repo root completes with no approval prompt. This is the one check that could not be run here.

## Related

Closes #250.

Follow-ups deliberately not folded in: a parsing hook for force-push, and the `green-gate.sh` `statusMessage` noise the issue itself excludes.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -16,6 +16,7 @@
       "Bash(git commit:*)",
       "Bash(git checkout:*)",
       "Bash(git restore:*)",
+      "Bash(git push:*)",
       "Bash(cd \"d:/Source/Repos/NetPace\":*)",
       "Bash(.specify/scripts/bash/:*)",
       "Bash(dotnet build:*)",
@@ -67,7 +68,6 @@
     "deny": [
       "Bash(sudo:*)",
       "Bash(su:*)",
-      "Bash(chmod:*)",
       "Bash(chown:*)",
       "Bash(curl:*)",
       "Bash(wget:*)",
@@ -84,7 +84,7 @@
       "Edit(.specify/scripts/bash/*.sh)"
     ],
     "ask": [
-      "Bash(git push:*)"
+      "Bash(chmod:*)"
     ],
     "defaultMode": "acceptEdits"
   },
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -73,12 +73,6 @@
       "Bash(wget:*)",
       "Bash(ssh:*)",
       "Bash(scp:*)",
-      "Read(**/.env)",
-      "Read(**/.env.*)",
-      "Read(**/secrets.*)",
-      "Read(**/.ssh/**)",
-      "Read(**/appsettings.*.json)",
-      "Read(**/appsettings.json)",
       "Edit(.claude/skills/speckit-*/SKILL.md)",
       "Edit(.specify/templates/*.md)",
       "Edit(.specify/scripts/bash/*.sh)"
--- a/docs/agentic-workflow-NetPace.md
+++ b/docs/agentic-workflow-NetPace.md
@@ -91,7 +91,7 @@ An `ask`-matched call **prompts** in an interactive session but is **silently de
 
 That asymmetry makes headless a permission oracle — run a workflow under `claude -p --dangerously-skip-permissions` and anything that would have prompted interactively comes back denied instead, with no human in the loop to mask it. Not CI-gateable: it needs the `claude` binary and an authenticated session.
 
-`Bash(rm:*)` and `Bash(rmdir:*)` came off the list for this reason ([CIR](change-intent-records/2026-09-04-rm-off-the-ask-list.md)), and `Bash(chmod:*)` moved onto it from `deny` — which buys an approval path interactively but not in a lane worker, where `ask` still denies silently ([CIR](change-intent-records/2026-09-04-push-allow-chmod-ask.md)).
+`Bash(rm:*)` and `Bash(rmdir:*)` came off the list for this reason ([CIR](change-intent-records/2026-09-04-rm-off-the-ask-list.md)), and `Bash(git push:*)` followed them off it into `allow` — that is what lets `/ship` reach `gh pr create` without stopping at its second-to-last step. `Bash(chmod:*)` moved the other way, from `deny` onto `ask`, which buys an approval path interactively but not in a lane worker, where `ask` still denies silently ([CIR](change-intent-records/2026-09-04-push-allow-chmod-ask.md)).
 
 ## Test-green gate & categories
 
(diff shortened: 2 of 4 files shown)


FrankRay78:
- Bugs or correctness issues

FrankRay78:
@claude Please re-review, given the most recent commit.

Issue #263, opened 2026-09-04
Gate the PR closing keyword in CI, conditioned on branch shape


FrankRay78:
## Problem

#248 made `/raise-pr` derive and verify a `Closes #<N>` keyword. That rule is prose read by a model, and nothing enforces it. The same step 6 that mandates the keyword also says *"Do not follow a fixed template. Decide which sections earn their place"* — so a future model, or a future edit to `raise-pr.md`, can silently drop back to the pre-#248 state. #248's own evidence table shows that regression is invisible until someone audits eight PRs by hand.

Constitution Principle I's carve-out does not end at recording RED/GREEN once:

> MUST still make the check repeatable. Where the tool can run in CI, add it there, **so the invariant is gated rather than verified once.**

## Why this was deferred out of #248

#248 argued no proportionate gate existed, on the grounds that "every PR body carries a closing keyword" would false-positive on legitimately issue-less PRs (dependabot bumps, `chore/normalise-cs-line-endings`). That reasoning was wrong, and is corrected here so it is not cited again: it defeats only the *unconditional* gate. A gate **conditioned on branch shape** has no such problem, because the condition is the shape itself rather than a maintained exception list.

Measured over the last 25 merged PRs, the separation is exact:

- All 6 branches matching `^feature/[0-9]+-` carry a closing keyword — so the gate never fails one.
- Every non-matching branch is exempt by construction, including `dependabot/*` and `chore/normalise-cs-line-endings`.
- `chore/234-format-ship-cadence` carries a keyword without matching, which is a false *negative* — the safe direction.

Zero false positives, no allowlist to maintain.

## Suggested shape

A step in the existing `.github/workflows/dotnet.yml` (already triggered on `pull_request`), roughly 20 lines against the ~20 lines of prose it guards — a ~1:1 ratio, so `feedback_guard_outgrows_the_fix` does not fire:

- Parse `^feature/([0-9]+)-` from `github.head_ref`; exit 0 when it does not match.
- Reuse the rule `raise-pr.md` step 6 already establishes: `gh issue view <N> --json state,url` and exit 0 unless the url is exactly this repo's `/issues/<N>` path — that rejects a PR at that number and an issue in another repo.
- Require a closing keyword for `#<N>` in the PR body, with a documented `No-close: #<N> — <reason>` escape for a branch that only partly implements its issue.
[...]

PR #264, opened 2026-09-04
PR Closing Keyword

### LINKED ISSUE #235: Harness sync: four portable items (piped gate exit codes, tooling install/report, decisiveness memory, transcript retention)

### LINKED ISSUE #248: /raise-pr: link the PR to its issue with a closing keyword
## Problem

`/raise-pr` never adds `Closes #<N>`, so a merged PR does not close the issue it implemented. Step 6 lists "**Related** — links to issues, prior PRs, or specs that give context", which invites a bare link and mandates nothing.

The result is that linkage depends on the invoker remembering to ask. Evidence from the last eight merged PRs — only #236 carries a closing keyword:

| PR | Closing keyword |
|---|---|
| #246 Ship Namespaced Reviewer Agents | no |
| #245 Upload Redirect 307 | no |
| #244 Capture Learnings Lenses | no |
| #241 Add /build | no |
| #240 Renormalise .cs line endings to LF | no |
| #238 Record Latency Flake Memory | no |
| #237 Profile XmlDoc File Sharing | no |
| #236 Move formatting to ship cadence | **yes** |

This surfaced on #247, where `Closes #235` only made it into the body because it was typed into `/ship`'s arguments by hand. `/build` → `/ship` → `/raise-pr` is meant to run unattended; a step that only works when a human remembers it is not automated.

## The trap to design around

The obvious mechanism — take the number off the branch name — is unsafe as stated, and `/raise-pr`'s own step 5 contains the counter-example: it documents `004-raise-pr-command` as a branch to strip a `NNN-` prefix from. **Spec-kit branches carry a sequence number, not an issue number.** Turning that into `Closes #4` would silently auto-close an unrelated issue on merge — a worse failure than the one being fixed, and one nobody notices until an issue is mysteriously closed.

So the number must be *verified*, not merely parsed.

## Suggested shape

- Derive a candidate issue number from the branch (`feature/<N>-<slug>` is what `/build` produces).
- Verify it with `gh issue view <N>` before using it: the issue must exist and be open. If it does not resolve, add no closing keyword rather than guessing.
- Accept an explicit override from the invoker, so a branch whose name carries no number can still be linked.
- Print what it decided, so `/ship`'s output shows whether the PR will close anything.

## Acceptance criteria

- [ ] A PR raised from a `/build`-created branch for an open issue carries `Closes #<N>` in its body without the invoker asking for it.
- [ ] A branch whose numeric prefix does not resolve to an open issue produces a PR with **no** closing keyword, and says so in the command's output.
- [ ] An explicitly supplied issue number is honoured over anything inferred from the branch.
- [ ] `/raise-pr`'s step 6 guidance states the rule, so the behaviour is documented where the body is composed.

## Related

- PR #247 — where this was noticed; `Closes #235` was supplied manually.

FrankRay78:
## Why

A merged PR did not close the issue it implemented unless the invoker remembered to type `Closes #<N>` by hand. `/raise-pr` step 6 only suggested "links to issues… that give context", which mandates nothing. #248 has the evidence: of the eight merged PRs before it, one carried a closing keyword.

The obvious mechanism — take the number off the branch name — is unsafe, and #248 says why: spec-kit branches carry a *sequence* number, not an issue number, so `004-raise-pr-command` would emit `Closes #4` and silently auto-close something unrelated. The number has to be **verified**, not merely parsed.

## What changes

- `/raise-pr` settles a closing keyword on every run: an invoker-supplied number wins the *sourcing*, the branch supplies a candidate otherwise, and either way `gh` verifies it before use.
- The outcome is three-way — linked, no issue, or unverifiable — and is reported, so the invoker can see what merging will close without opening GitHub.
- `/ship` forwards an issue number through to `/raise-pr` and surfaces the outcome in its final report.
- `/build`'s READY message no longer tells the invoker to add the keyword by hand.

This PR is its own first exercise of the feature: the `Closes #248` below was derived from the branch name and verified, not typed.

**Verification proves existence and openness, never identity.** It confirms `<N>` is an open issue in this repo — never that it is *this branch's* issue. That gap is held narrow by restricting inference to the `/build` branch shape rather than by the check itself, and the prose says so rather than implying the check is stronger than it is.

**A single URL-equality test replaces three separate checks.** Requiring `url` to equal this repo's `/issues/<N>` rejects, in one comparison, a pull request at that number, an issue in another repo, and a transferred issue that redirected. Each was reproduced against the live tool first:

- `gh issue view 4` returns `{"state":"MERGED","url":".../pull/4"}` — it resolves a *PR* rather than failing, so a state-only check would have passed for an open PR. This one was found by exercising the rule, not by reading it.
- `gh issue view https://github.com/cli/cli/issues/14353` run from inside NetPace returns OPEN at an `/issues/` path — a pasted cross-repo link verified clean and would have emitted a `Closes` meaning something else here.

**"No such issue" and "couldn't check" both exit 1.** A nonexistent issue and an expired token are indistinguishable by exit code, so the rule reads stderr instead: `Could not resolve to an issue or pull request` is an answer, `HTTP 401: Bad credentials` is a broken check. Folding them together would report a fixable auth failure as "branch carries no issue number" and send the invoker looking in the wrong place.

**Branch inference deliberately declines more than it accepts.** A bare `NNN-` spec-kit branch yields *no* candidate rather than one the checker is trusted to reject, and the number is only ever read from the start of the segment — `feature/net10-upgrade` must not yield `10`. False negatives here are the safe direction; the cost is a missing link, not a wrongly-closed issue.

**A failed lookup is an abstain, not a `/ship` failure.** `/ship`'s stop-on-failure rule is global, so the exception is named explicitly to stop the two documents disagreeing.

## Corrected reasoning, recorded so it is not cited again

The `/build` report justified adding no CI gate on the grounds that the only gate shape would false-positive on issue-less PRs. That was wrong — it defeats only the *unconditional* gate. A gate conditioned on branch shape has no such problem: across the last 25 merged PRs, all 6 `feature/<N>-` branches carry a keyword and every other branch is exempt by construction, dependabot included. Zero false positives. Opened as **#263** rather than built here, because its `No-close:` escape hatch is new repo policy and a second mission.

Also worth stating plainly: the grep showing the old prompt mandated nothing is a diff summary, not RED evidence — the string matches because it was typed in. The real RED→GREEN is the `gh` probe sequence above, where a candidate rule was executed, produced a wrong answer against a real input, and was strengthened until all cases separated.

## How to verify

- [ ] `gh issue view 4 --json state,url` — confirm it resolves a *pull request*, which is what makes the URL check load-bearing rather than belt-and-braces.
- [ ] `gh issue view https://github.com/cli/cli/issues/14353 --json state,url` from this repo — confirm a cross-repo URL still returns OPEN at an `/issues/` path.
- [ ] `GH_TOKEN=invalid gh issue view 248` vs `gh issue view 999999` — confirm both exit 1 with different stderr, so only stderr can separate them.
- [ ] Read `raise-pr.md` step 6 as an agent would: does any path reach `Closes` without passing all three conditions?

## Related

Closes #248

- #263 — follow-up: gate this invariant in CI, conditioned on branch shape.
- PR #247 — where the missing linkage was noticed; `Closes #235` was supplied by hand.

https://claude.ai/code/session_01SxP7idFUPigWdvUMPtuFuX

### DIFF
--- a/.claude/commands/build.md
+++ b/.claude/commands/build.md
@@ -97,5 +97,5 @@ Two named exceptions, because `CLAUDE.md` requires discussion for them:
 - **What you changed**, at a behaviour level, and how each acceptance criterion is met.
 - **Any assumption** you made on an ambiguous point, any public-API change, and anything you deliberately left out of scope.
 - Then exactly one of:
-  - `READY branch=<branch>` — every criterion implemented, whole suite green, everything committed, tree clean. Follow with: "Run `/ship` to format, gate, review and raise the PR — `/raise-pr` derives `Closes #<N>` from this branch name and verifies it, so the issue closes on merge."
+  - `READY branch=<branch>` — every criterion implemented, whole suite green, everything committed, tree clean. Follow with: "Run `/ship` to format, gate, review and raise the PR — `/raise-pr` derives `Closes #<N>` from this branch name and verifies it before use, then reports what it settled; check that line to confirm the link was made."
   - `FAILED reason=<short reason>` — you could not reach that state. Report the wall you actually hit, discovered by working: the criteria conflict, they do not determine the design, the change is larger than they describe. Do not fabricate READY.
--- a/.claude/commands/build.md
+++ b/.claude/commands/build.md
@@ -97,5 +97,5 @@ Two named exceptions, because `CLAUDE.md` requires discussion for them:
 - **What you changed**, at a behaviour level, and how each acceptance criterion is met.
 - **Any assumption** you made on an ambiguous point, any public-API change, and anything you deliberately left out of scope.
 - Then exactly one of:
-  - `READY branch=<branch>` — every criterion implemented, whole suite green, everything committed, tree clean. Follow with: "Run `/ship` to format, gate, review and raise the PR — and add `Closes #<N>` to the PR body if you want the issue to close on merge."
+  - `READY branch=<branch>` — every criterion implemented, whole suite green, everything committed, tree clean. Follow with: "Run `/ship` to format, gate, review and raise the PR — `/raise-pr` derives `Closes #<N>` from this branch name and verifies it, so the issue closes on merge."
   - `FAILED reason=<short reason>` — you could not reach that state. Report the wall you actually hit, discovered by working: the criteria conflict, they do not determine the design, the change is larger than they describe. Do not fabricate READY.
--- a/.claude/commands/raise-pr.md
+++ b/.claude/commands/raise-pr.md
@@ -4,6 +4,14 @@ description: Raise a PR for the current feature branch — cleans up the spec fo
 
 Read `CLAUDE.md` for project context before proceeding.
 
+## User Input
+
+```text
+$ARGUMENTS
+```
+
+Optionally a GitHub issue number — bare (`248`), hashed (`#248`), or a full issue URL — naming the issue this PR closes. Empty is the normal case and is **not** a prompt: step 6 infers the issue from the branch instead. `/raise-pr` never asks the invoker for anything, so it stays composable by `/ship` with no interactive gate.
+
 ## Steps
 
 1. **Get branch name**: Run `git rev-parse --abbrev-ref HEAD`. If the result is `main`, stop immediately and output: "Run /raise-pr from a feature branch, not main."
@@ -26,6 +34,16 @@ Read `CLAUDE.md` for project context before proceeding.
 
 6. **Compose PR body**: Write a description sized to the change. **Do not follow a fixed template.** Decide which sections earn their place based on what's actually in this PR.
 
+   **First, settle the closing keyword.** A merged PR must close the issue it implemented without the invoker having to remember to ask for it, so decide this every time — not only when someone thinks to mention an issue:
+
+   - **An issue number supplied by the invoker wins.** Take it from the User Input above (bare, `#N`, or an issue URL) and use it directly — no branch inference, no second-guessing.
+   - **Otherwise infer a candidate from the branch name.** `feature/<N>-<slug>` is what `/build` produces, so the leading number on the segment after the prefix is the candidate.
+   - **Verify the candidate before using it — never trust the parse.** Run `gh issue view <N> --json number,state,title,url`. Use `Closes #<N>` only if all three hold: the lookup succeeded, `state` is `OPEN`, and `url` is an `/issues/<N>` path. Otherwise add **no** closing keyword and carry on composing the body.
+   - Verifying is the point of the rule, not ceremony around it: spec-kit branches carry a *sequence* number, not an issue number — `004-raise-pr-command` is the very branch step 5 strips `004-` from — so an unverified parse would emit `Closes #4` and silently auto-close an unrelated issue the moment the PR merged. That is a worse failure than the missing link this rule exists to fix, and nobody notices it until an issue is mysteriously closed.
+   - The `url` check is not belt-and-braces. GitHub numbers issues and pull requests from one sequence, and `gh issue view <N>` resolves a *pull request* at that number rather than failing: on this repo `gh issue view 4` returns `{"number":4,"state":"MERGED","url":".../pull/4"}`. A state-only check therefore passes for an open PR and would emit a `Closes` aimed at a pull request. The `/issues/` path is what separates the two.
+   - The keyword goes on its own line in **Related**. The PR body is the *only* place a closing keyword belongs — never a commit message, which would close the issue as soon as it reached `main`, ahead of review.
+   - Whichever way it lands, step 10 reports it.
+
    **Guiding principle — don't duplicate things a reviewer can get elsewhere.** GitHub already shows:
    - the list of changed files and their diffs
    - the commit list and commit messages
@@ -40,7 +58,7 @@ Read `CLAUDE.md` for project context before proceeding.
    - **What changes** — a short summary of observable changes. For single-theme PRs, 1–3 bullets. For multi-theme PRs, group by theme rather than by file. Describe *what* changed at a behaviour level, not which files were touched.
    - **Non-obvious things a reviewer should know** — caveats, deliberate trade-offs, hidden invariants, anything in the diff that looks weird but is intentional. Omit if there are none.
    - **How to verify** — a markdown checklist of concrete actions a reviewer can perform. Derive items from what actually changed. Omit for pure-documentation PRs where there's nothing functional to exercise.
-   - **Related** — links to issues, prior PRs, or specs that give context. Omit if none.
+   - **Related** — the `Closes #<N>` line settled above, when one was settled, plus links to prior PRs, related issues, or specs that give context. Omit only if there is neither.
 
    A tiny PR may collapse to two sections. A sprawling PR may use all five. Match the size of the body to the size of the change.
 
@@ -60,4 +78,4 @@ Read `CLAUDE.md` for project context before proceeding.
    End the review with a recommendation whether to merge the PR, and if not, what you suggest needs addressing first.
    ```
 
-10. **Output result**: Print the PR URL. If step 4 deleted a spec folder, also print `Deleted spec folder: <FEATURE_DIR>`.
+10. **Output result**: Print the PR URL. If step 4 deleted a spec folder, also print `Deleted spec folder: <FEATURE_DIR>`. Always print what step 6 settled — `Closes #<N> — <issue title>`, or, when it settled nothing, that fact and the reason (`No closing keyword: branch carries no issue number`; `No closing keyword: #4 is a pull request, not an open issue`) — so the invoker can see from the output alone whether merging this PR will close anything, and what.
--- a/.claude/commands/raise-pr.md
+++ b/.claude/commands/raise-pr.md
@@ -10,7 +10,7 @@ Read `CLAUDE.md` for project context before proceeding.
 $ARGUMENTS
 ```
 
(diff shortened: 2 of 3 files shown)


FrankRay78:
- Bugs or correctness issues

PR #269, opened 2026-09-05
Context Mode Counters

### LINKED ISSUE #262: plugin-report.sh: context-mode's PERFORMANCE row points at ctx_stats instead of reporting counters
`scripts/plugin-report.sh` reports live counters for two of its three token/context tools and a pointer for the third.

## The gap

rtk and read-once both have a shell subcommand the script can invoke, so their PERFORMANCE rows carry real figures. context-mode's row does not:

```sh
# line 270
if [ $rtk_inst -eq "$YES" ]; then perf rtk rtk gain; else echo "rtk: n/a — not installed"; fi
# line 276
if [ -x "$ro_dir/read-once" ]; then perf read-once "$ro_dir/read-once" stats; else echo "read-once: n/a — not installed"; fi
# line 277
if [ $cm_inst -eq "$YES" ]; then echo "context-mode: installed — counters are MCP-only; run ctx_stats in-session"; else ...
```

The section header promises "Live counters". The context-mode row instead hands the reader a chore in a different place — open a Claude session, run `ctx_stats`, then mentally join the answer back to a report that has already scrolled past. A report whose whole product is a truthful answer should go and get the answer.

`backlog/dispatcher/lane-peek.sh` (a sibling project, not this repo) already establishes that a shell script can spawn `claude -p`, capture stdout, and fold the result back into its own output — including the argument-order and `--system-prompt` details needed to make a headless summariser behave. That was the assumed fix here.

## The premise is wrong, and the cheap fix was missed

The counters are *not* MCP-only. Verified against context-mode 1.0.169:

**The stats are plain JSON on disk**, one file per PID, under the storage root `context-mode doctor` reports:

```
$ cat ~/.claude/context-mode/sessions/stats-pid-*.json
{"schemaVersion":2,"version":"1.0.169","updated_at":…,"total_calls":0,"bytes_returned":0,
 "kept_out":0,"reduction_pct":0,"tokens_saved":0,"dollars_saved_session":0,
 "tokens_saved_lifetime":270080,"dollars_saved_lifetime":1.35,"by_tool":{}}
```

**A one-line summary is available from the shipped CLI:**

```
$ node <installPath>/cli.bundle.mjs statusline
context-mode  ●  saves ~98% of context window
```

There is no `stats` subcommand on that CLI (`index`, `search`, `doctor`, `upgrade`, `hook`, `statusline` only), so the *formatted* `ctx_stats` output really is MCP-only — but the underlying numbers are not.

Note also that the plugin ships a `stats.json` at its install root. It holds npm download counts for the README badge (`"label":"users","message":"535k+"`), not session counters. It is a trap for exactly this change; do not wire it up.

## Choosing between them

Three ways to fill the row, cheapest first:

1. **Read the JSON store** — instant, no subprocess, no network, no spend, and it keeps the script's "starts nothing" promise intact. Costs a dependency on a private on-disk shape, though the files carry `schemaVersion`, so a bumped version can be reported as `unknown` rather than silently misread.
2. **Call `context-mode statusline`** — a supported CLI surface, so it will not drift. But it yields one summary sentence and no token or spend figures, which is thinner than what the rtk and read-once rows print.
3. **Spawn `claude -p` with `ctx_stats` allowed** — returns the canonical, fully formatted output. It also costs seconds and real money per run, needs the network, and starts a model from a script whose header block says in capitals that it installs nothing, edits nothing and *starts nothing*. That header is a deliberate promise about a script people are asked to run on an unfamiliar box.

Recommendation: **(1), falling back to (2)**. It is the only option that leaves the READ-ONLY contract standing, and it is the one that actually produces figures comparable to the other two rows. Option 3 is a large price — spend, latency and a broken invariant — for formatting.

If option 3 is chosen anyway, the header's READ-ONLY paragraph must be amended in the same change to say the script starts a model, and the call must be opt-out (as `lane-peek.sh` does with `--no-ai`), never opt-in-by-default on a script advertised as inert.

## Second defect in the same row

`reachable=` for context-mode is hardcoded:

```sh
# line 162
cm_reach=$NA   # counters and health live behind MCP (ctx_doctor); nothing to probe from a shell
```

The comment is out of date: `context-mode doctor` runs from a shell and reports platform, storage paths, hooks and FTS5 health. context-mode is the only row claiming `n/a` for a reason that no longer holds — `pr-review-toolkit`'s `n/a` is genuine, since its agents are resolved by the harness.

The binary is not on `PATH` (the plugin exposes `bin: {"context-mode": "./cli.bundle.mjs"}` but is not npm-linked), so any probe must resolve `installPath` from `installed_plugins.json` — which the script already does in `plugin_installed()`.

## Acceptance criteria

- [ ] The PERFORMANCE section reports context-mode's usage counters, or context-mode's own "no data yet" equivalent when nothing has been captured — not an instruction for the reader to go and run something elsewhere.
- [ ] When the counters cannot be read, the row says so as an unreadable probe, and never as a zero or a healthy-looking figure ("I could not look" is not "no").
- [ ] The TOOLING section's `reachable=` value for context-mode reflects whether context-mode passes its own health check, and reads `n/a` only if no such check is reachable.
- [ ] The script's stated READ-ONLY contract and the behaviour of the code agree with each other after the change.
- [ ] TOOLING / CONFIG / HOOKS output remains free of timestamps and machine-specific paths, so two boxes still diff cleanly.

FrankRay78:
## Why

The PERFORMANCE section promised live counters but handed the reader a chore for context-mode: open a Claude session, run `ctx_stats`, then join the answer back to a report that has already scrolled past. The same row's `reachable=n/a` was stale too — `context-mode doctor` runs perfectly well from a shell.

**The issue's premise turned out to be wrong, and that shaped the whole branch.** #262 established that context-mode writes per-PID `stats-pid-*.json` sidecars and recommended simply reading them. That was implemented first. It is wrong: context-mode's own `bin/statusline.mjs` records those sidecars as legacy and *"no longer the source of truth"* (eventually-consistent, PID-scoped), and on a live box they disagreed with `ctx_stats` by ~80% — $2.01 across 5 sessions where `ctx_stats` reported $3.65 across 33 conversations. A `schemaVersion` guard cannot catch that, because the risk is not shape drift but **source** drift: a frozen legacy store passes the guard and prints confident, wrong figures forever. Reading them would have reproduced, in a new place, the exact fault the issue was raised to fix.

So the row now asks context-mode instead.

## What changes

- **PERFORMANCE reports the one figure that is genuinely about this box**, obtained by spawning a headless `claude -p` with only the `ctx_stats` MCP tool allowed — the same handler the tool answers with itself.
- **`reachable=` for context-mode runs `context-mode doctor`** instead of being hardcoded `n/a`.
- **The script no longer calls itself READ-ONLY.** It now states what it spends, writes and reaches over the network. That includes an npm-registry lookup inside `doctor` which an earlier draft of this very branch denied while introducing it.
- **The verdict mapping is fixed on every row, not just the new one.** `read-once` and `rtk` assigned a raw `$?` to their verdict, so a probe exiting 3 printed `reachable=n/a` — which this report defines as *"there is no probe for this cell"*, asserting none was attempted when one ran and failed.

- **Only one line of `ctx_stats`' reply is reported, and that was measured rather than assumed.** The same probe run from this repo and from the Claude home returns an identical section-3 "All your work" total but a *different* section-4 dollar figure — $3.84 against $2.80, each stable on repeat. Section 4 reads like a box-wide bottom line and is scoped to wherever the probe ran, so it is excluded.
- **The probe is confined, deliberately.** `--allowedTools` is a permission *grant*, not a whitelist — `Read`, `Glob`, `Grep` and `Task` stay auto-approved unless denied. A model asked for context-mode's counters with `ctx_stats` unavailable would go looking, and find the legacy sidecars this branch exists to stop trusting.
- **It runs from a fixed neutral directory**, so it does not execute the invoking repo's hooks (a blocking `Stop` hook would force another turn and could make the captured reply the model's answer to the hook). Fixed rather than a fresh temp dir, because context-mode counts each distinct directory as a project — a throwaway per run inflated the project count in the very figure being reported (observed going 10 → 12 during testing).
- **An unrecognised reply is never printed as a reading.** `claude -p` does not guarantee a verbatim echo; it may paraphrase, refuse, report a usage limit or be truncated, and all of those exit 0.
- **Author override, flagged for scrutiny.** #262 made an opt-out flag (a `--no-ai`) a *condition* of taking this route. The author chose always-on with no flag — the header discloses the cost instead. Recorded in the CIR so the departure is visible rather than lost.
- **The empty-`TIMEOUT` fix is a cross-platform bug, not tidiness.** The bare array expansion is a fatal unbound-variable error under `set -u` on bash < 4.4, and macOS ships bash 3.2 *and* no `timeout(1)` — the same box — so the report aborted mid-TOOLING there.
- **No committed test file, by prior decision.** #235's confirmed decisions place this script at `scripts/` with no `*.tests.sh` matrix — it is a one-off manually-run script, not a gate. The RED/GREEN harness (54 assertions) lives in the gitignored scratch tree, as it did for #252.

## How to verify

- [ ] `bash scripts/plugin-report.sh` — the context-mode PERFORMANCE row carries a box-wide figure labelled `(via ctx_stats, headless probe)`, and TOOLING shows `reachable=yes`. Takes ~16s; it starts a model, which is the point of the header rewrite.
- [ ] Read the header block against the code. It is the document someone trusts before running this on an unfamiliar box, and an earlier draft of this branch got it wrong — that is why it is called out here.
- [ ] `PATH=/usr/bin:/bin bash scripts/plugin-report.sh` — with `claude` unreachable the row must read `unknown`, never a zero or a figure.
- [ ] Confirm `docs/agentic-workflow-NetPace.md` and `.claude/commands/install-harness-tooling.md` describe the same contract as the header — all three drifted before.
- [ ] `dotnet build ./src && dotnet test ./src` — unaffected by this branch (648 tests, 0 skipped), but the gate ran throughout.

## Related

Closes #262

Follows #252, which fixed the `read-once` row in the same section.

https://claude.ai/code/session_01Df96UWHo1HjY1EB1aH84gZ


FrankRay78:

- Bugs or correctness issues

FrankRay78:
FYI. Running the command in claude produced a much richer output, which really should be included in the script output IMHO, see: 

image

FrankRay78:
Pushed `50436a4` in response to the feedback above — the full `ctx_stats` reply is now printed, and the probe runs in the repo rather than a neutral directory.

**What changed and why:**

- **Slicing is gone.** Keeping only the one line proven to be box-wide also discarded the opening headline totals (`Across N days you ran M conversations … about K KB every single day`), which are box-wide and among the most useful lines in the reply. That was a mistake; the reply is printed whole.
- **The probe now runs in the repo.** The reply is scoped to the directory the probe runs in — that was the original reason for excluding the dollar figure. Running in the repo makes that scope the right one: section 1 carries this repo's real figures (763 KB → 195.4K tokens, 100% kept out) instead of the empty throwaway session a neutral directory produced.
- **A side benefit:** no phantom project. A fresh directory per run added one to context-mode's project count every time — observed pushing it from 10 to 12 during development, inflating a number the row itself reports.
- **The accepted cost, now stated rather than denied:** the nested session runs this repo's hooks. The previous header claimed the opposite. The payload validation is what keeps a hook-mangled reply out of the row, and the row stays labelled `(via ctx_stats, headless probe)` so "this chat" in the reply reads as the probe's session.

**On the residual gap you named** — `--disallowedTools` enumerating only built-in tool names, so globally-enabled MCP tools from other plugins are neither granted nor denied: that is correct and I had not considered it. It is narrower than it was, since the probe now runs in this repo where `.claude/settings.json` governs, but "confined to `ctx_stats`" is still only true of the built-in tool surface. Worth a follow-up issue rather than widening this PR — an allow-list approach would need a way to enumerate what a box has enabled, which the script cannot see.

Scratch harness: 54 assertions, deterministic. .NET suite unaffected and green (648 tests, 0 skipped, 0 warnings).

---

@claude Please review the changes pushed since your last review — specifically commit `50436a4` (`git diff 687a28e..50436a4`), in the context of the full PR. Focus on:

- Whether printing the reply whole reintroduces any way for non-counter text to be presented as a reading
- Whether running the nested session inside this repo (rather than a neutral directory) creates problems the previous design avoided — hook execution, recursion, or scope confusion in the reported figures
- Whether the header block and the CIR now accurately describe what the code does, given both were wrong on this exact point one commit ago
- Anything in the diff that contradicts the reasoning recorded in the CIR

End with a recommendation whether to merge.

FrankRay78:
Both items addressed in `33b11dc`.

**1. Stale CIR paragraph — fixed.** The "Chosen" paragraph still described the sliced-output, neutral-directory design and contradicted two paragraphs below it. Corrected to point at the scope reasoning rather than restate a superseded version of it.

**2. The validation overclaim — closed, not softened.** You were right that "the validation keeps a hook-mangled reply out of the row" promised more than substring-presence delivers, and right about the mechanism: marker presence proves only that the genuine output is in there *somewhere*, which is precisely the dangerous case, since a Stop-hook-forced extra turn leaves the real counters intact and appends after them.

I tried to remove the cause first. `claude --restricted` looked ideal — it ignores project and local settings, so this repo's hooks would not run — but it ignores *user* settings too, which unloads the context-mode plugin and makes `ctx_stats` unavailable. Not usable. That attempt did usefully confirm the gate works end-to-end: the model returned prose explaining the tool was missing, exited 0, and the row reported `unknown`.

So the fix is an end anchor. A genuine reply terminates at `ctx_stats`' footer, a bare version line:

```
  Locale en-US · timezone Europe/Berlin · pricing examples for illustration only.

  v1.0.169
```

The reply must now end there as well as carry both section markers. Trailing text — the shape a hook-forced turn produces — fails validation, and the row reports `unknown` and shows the reply under the `NOT counters` label. If that footer ever changes, every run reports `unknown` and prints the raw reply: a visible failure rather than a silently wrong figure, which is the trade this script makes everywhere else.

The comment and the CIR now both describe what is actually checked, and say which half is load-bearing and why.

**Harness:** 57 assertions, deterministic over repeated runs. The new one pins exactly your scenario — genuine counters with commentary appended after the footer, which only the end anchor rejects. Verified a real reply still validates cleanly. .NET suite green (648 tests, 0 skipped, 0 warnings).

---

@claude Please review `33b11dc` (`git diff 50436a4..33b11dc`). Specifically:

- Does the end-anchor check actually close the gap you identified, or can trailing text still reach the row by another route (text prepended before the counters, a reply with the footer appearing mid-document, multiple tool calls in one reply)?
- Is the version-line footer a safe thing to depend on, given the failure mode is a permanent `unknown` if it changes? Is that trade stated accurately in the comment and CIR?
- Do the comment at `cm_reply_is_stats` and the CIR now describe exactly what the code checks, with no remaining overclaim?
- Anything else in the diff worth raising before merge.

End with a recommendation whether to merge.

PR #272, opened 2026-09-07
Verify Split

FrankRay78:
## Why

`/ship` did two separable jobs behind one invocation: it verified the branch, then raised the PR. Raising the PR is the only irreversible, outward-facing act in the chain — everything before it is safe to re-run, and pushing a branch and opening a PR is not. Splitting there is what makes the rest of the chain freely repeatable, and it turns the SDLC commands into stages that can each be run alone, by hand or by a script.

After the cut the command formats, tests, reviews and commits. It no longer ships anything, so the name was renamed to match what it actually guarantees.

## What changes

**The command.** `.claude/commands/ship.md` → `verify.md`, with its PR step removed. `/verify` now ends at a green, reviewed, fully-committed branch with a clean tree — already exactly `/raise-pr`'s entry condition, so the two compose by hand with no adapter step. Its closing report now gives an explicit `VERIFIED`/`FAILED` verdict and must *name* any finding it deferred.

**Three rationale passages were re-argued, not renamed.** Each justified itself by the fact that `/ship` called `/raise-pr`, so a verbatim swap would have left all three wrong:

- The step-0 precondition guard lost its "don't let `/raise-pr`'s late branch check be the first line of defence" argument, and now stands on cost alone.
- The no-docs-only-skip argument **inverted**. It used to lean on the `gh pr create` hook re-running the suite later in the same command; that hook now sits in a stage that may never run, so the verify-time run became the load-bearing one rather than the redundant one.
- Review B moved outside the gate entirely, since `/raise-pr` requests it.

**Vocabulary, not just command references.** "The ship gate" and "formatting at ship cadence" are renamed throughout, including in the otherwise project-agnostic `docs/agentic-workflow.md`, so a reader meeting the term anywhere is pointed at a command that exists.

## Non-obvious things a reviewer should know

**The rename is two commits on purpose.** A single commit rewrote enough of the file that git would not pair it even at `-M20%`, defeating the "history follows" intent. It is split into a byte-identical `git mv` and then the rewrite, so `git log --follow .claude/commands/verify.md` reaches back through #246, #236 and #225. Review the second commit for content; the first is a pure move.

[...]

PR #273, opened 2026-09-07
Decouple Verify Step Numbers

FrankRay78:
## Why

Thirteen references across six files encoded `/verify`'s step *numbering* from outside `verify.md`. Every one was correct, and every one would have been silently falsified by inserting a step into that command — a reader following "`/verify` step 2" to the review step would land somewhere else, with nothing failing to say so.

This is the coupling constitution Principle IX describes: the reference was pinned to a mechanism (the ordinal) rather than to the observable thing it names (the formatting step, the suite gate, the review step). Renaming `/ship` to `/verify` demonstrated that step numbers in this file do move.

## What changes

Each reference now leads with the step's name and keeps the number as a parenthetical — "`/verify`'s suite gate (step 1b)", "`/verify`'s clean-context review step (step 2)". The name survives renumbering; the number stays for anyone who wants to jump straight to it. Nothing is removed, and no behaviour changes.

The two references *inside* `verify.md` are deliberately left as bare numbers. They are self-consistent, and anyone renumbering the file is already reading it — those fail loudly rather than silently.

**The count in this PR was wrong four times before it was right: 6 → 9 → 10 → 13.** Every survey but the last anchored its regex on proximity to a `/verify` token, so each successive grep re-inherited the same blind spot and "confirmed" the previous undercount. What it kept missing were references phrased with a bare capitalised "Step" at sentence start — including three in `agentic-workflow-NetPace.md:82-83`, in the very section this PR body called its most exposed target.

The final sweep matches **every** step-number token in the repo with no proximity condition, then classifies each by hand. Thirteen refer to `/verify`'s steps; the rest belong to other documents' own numbering (`raise-pr`, the speckit commands, `wsl-claude-sandbox`, `agentic-workflow.md`'s command list) and are deliberately untouched.

This is the same failure mode as the `/ship` sweep that missed "ship command" in #272 — a proximity-anchored grep reported clean and was believed. Reviewers should treat the enumeration below as the claim under test, not as established.

**No gate was added for this.** A hook asserting "no bare step-number reference to `/verify`" would be larger and buggier than the ten edits it polices, which is the wrong trade at this size.

## How to verify

- [ ] `grep -rnoiE '\bstep[- ][0-9]+[a-z]?\b' .` — every hit is either one of the 13 `/verify` references in `(step N)` form, or another document's own numbering. No proximity filter; that is what hid three of them.
- [ ] Every reference reads as a step *name* with the number in parentheses; no bare "`/verify` step N" survives outside `verify.md` and `build.md`'s self-references.
- [ ] Each number still points at the right step in `.claude/commands/verify.md` — 0 preconditions, 1a format, 1b suite, 2 review, 3 re-verify and commit.

## Related

Follow-up to #272, which split `/ship` into `/verify` and a manual `/raise-pr`. This cleanup was identified during that PR's clean-context review and deliberately deferred as out of scope, then carried in its PR body; raised here directly rather than via an issue.


FrankRay78:

- Bugs or correctness issues

FrankRay78:
@claude Please re-review this PR, focusing specifically on **consistency**. Two commits have landed since your last review (c88662b, then 93e7229 which supersedes it), so please review the branch as it now stands rather than the earlier state.

The PR's claim is that every reference to one of `/verify`'s steps now names the step and carries the number as a uniform `(step N)` parenthetical. Please verify that claim holds, and check:

- **Uniformity of the parenthetical.** Is it `(step N)` in every instance, with no stragglers in an older form — a bare `step N`, a `(N)` without the word, a comma instead of parentheses, or a `step (step N)` stutter?
- **Completeness.** Is the set actually complete? The count was revised twice during this work (six → nine → ten) because grep-based surveys under-reported: one reference sat outside a proximity window, and three others were never in the original list. Please enumerate independently rather than trusting the PR body's count, and include `.claude/`, `docs/`, `scripts/` and any hook comments.
- **Step-name accuracy.** Each number should point at the step the name claims: 0 preconditions, 1a formatting, 1b the suite gate, 2 the clean-context review, 3 re-verify and commit. Check each against `.claude/commands/verify.md`.
- **Naming consistency.** The same step should be described the same way across files. Note `step 2` is called "clean-context review" in two places and just "review" in a third — is that acceptable variation in context, or should it be uniform too?
- **The deliberate exception.** The two references *inside* `verify.md` are intentionally left as bare numbers, on the grounds that they are self-consistent and fail loudly. Is that the right call, or should they follow the same convention?
- **House style.** `verify.md:9` already used `(step 1a)` before this PR. Does the new wording match the surrounding register, or does anything read as awkward — particularly "formatting pass" and "committing the fixes", which were reworded to avoid the stutter?

Also flag any prose that hard-wraps at a fixed column, which CLAUDE.md forbids.

End with a recommendation whether to merge, and if not, exactly what needs addressing.

FrankRay78:
@claude Your finding was correct and is now fixed in f011fb6. Three bare references at `docs/agentic-workflow-NetPace.md:82-83` — "Step 1a runs...", "step 3's clean-tree invariant", "Step 1b is..." — plus the `rm-off-the-ask-list.md:6` naming straggler you asked to bundle, which now reads "clean-context review" like the other two sites.

**The true set is 13, not ten.** The PR body has been corrected.

Root cause, stated plainly because it matters more than the fix: the count was wrong four times running — 6 → 9 → 10 → 13 — and every survey but the last anchored its regex on proximity to a `/verify` token. The three you found begin with a bare capitalised "Step" at sentence start with no such token nearby, so each successive grep re-inherited the same blind spot and appeared to confirm the previous undercount. A clean result from a filtered search was repeatedly mistaken for a clean result.

The final sweep drops the proximity condition entirely: it matches every `step N` token in the repository, then classifies each by hand. 13 refer to `/verify`'s steps and are now uniformly `(step N)`. The rest belong to other documents' own numbering — `raise-pr`, the speckit commands, `wsl-claude-sandbox`, and `agentic-workflow.md`'s command list — and are untouched.

Two things I did **not** change, and would like you to rule on rather than assume:

1. **Self-references stay bare.** `build.md:39,70,71,76` and `verify.md:57,66` point at their own files' steps. They are self-consistent and anyone renumbering those files is reading them, so they fail loudly rather than silently. If you think uniformity should win over that argument, say so.
2. **No gate was added.** A hook policing this would be larger and buggier than 13 prose edits. But given a manual enumeration was wrong four times, there is a real argument that this specific invariant is exactly the kind that should not be left to human greps. I lean against on proportionality; I would rather you challenge that than agree by default.

Please re-check the enumeration independently — do not trust the 13. Suite is green (648 passed, 0 skipped) and both hook matrices pass, though nothing here touches C#.

Issue #271, opened 2026-09-05
Automated pre-specification review on labelled issues


FrankRay78:
## Summary

Applying a designated label to a GitHub issue produces a NetPace pre-specification gap analysis as a comment on that issue, with no terminal involved. The author answers the questions inline and continues with the existing confirm step unchanged.

## Motivation

`/speckit.reviewissue` is the pre-specification gate: it reads an unrefined issue, cross-references it against the codebase, and posts a numbered list of gaps for the author to answer. Today it runs only from a developer machine.

That makes the first step of a three-step loop the one that gates the other two. An issue raised away from the desk sits untouched until the author sits down — and the review is precisely the step that benefits from being done ahead of time, because it is the one that carries the codebase grounding and therefore the wait. Answering the questions and running `/speckit.confirmissue` are quick by comparison, and both are fine to do at the machine.

The existing `@claude` automation ([.github/workflows/claude.yml](.github/workflows/claude.yml)) already proves the shape works for pull requests — `/raise-pr` posts a review request and Claude answers on the PR. This issue asks for the equivalent on the issue side of the workflow, where the payoff is a review waiting rather than a review requested.

## Users & jobs

**Repository maintainer, away from the machine.** Raises an issue from a phone or tablet, marks it as wanting a review, and wants the gap analysis waiting when they next sit down — specific to this codebase, not generic advice.

**Repository maintainer, at the machine.** Wants the automated review to be a first-class citizen of the existing pipeline: answerable inline, and foldable into the issue body by the existing confirm step with no manual repair.

## Capability

**Triggering.** Marking an issue as wanting a review causes a review to be produced for that issue. Marking is deliberate — issues are not reviewed merely by existing, because bugs and trivial changes do not warrant the overhead. Only people who can mark an issue in this repository can cause a review to run.

**The review itself.** The result appears as a single comment on the issue, matching what a locally-run review produces: numbered gaps, a recommendation per gap, and a slot for the author's answer. It is grounded in the actual codebase — it cites real paths, real conventions and real neighbouring code, rather than generic observations that would apply to any repository.

**Continuity with the existing loop.** The author answers the questions inline on the comment, then runs the existing confirm step, which folds the answers into the issue body as confirmed decisions. No manual repair, re-formatting, or re-posting is needed in between.

**Self-documenting constraint.** A review produced this way is a first pass only. The comment itself tells the reader what to do when a question needs expanding, so the constraint is discoverable at the point of use rather than only in a workflow file.

[...]

Issue #260, opened 2026-09-04
Add /study — record what surprised a build, so the harness can be improved from evidence


FrankRay78:
## Summary

Add a `/study` command that records what *surprised* a piece of work — a mid-flight redesign, a reviewer finding that had to be actioned, an acceptance criterion that turned out ambiguous — as a short, classified table in `docs/study/<issue>.md`. Run at any point, it complements `/build` and `/verify` without being wired into either. Over time the accumulated files answer a question no single PR can: *which part of the harness keeps costing us, and therefore what to fix.*

## Motivation

`/build` and `/verify` now carry a piece of work from a GitHub issue to a reviewed, test-green branch. What neither leaves behind is any record of where the work went sideways on the way. A reviewer finding gets fixed and disappears into a commit; an ambiguous criterion gets interpreted and never mentioned again. Each is invisible individually, and collectively they are the only evidence of whether the harness itself needs work.

A sister project running the same command set under an autonomous dispatcher has been keeping exactly this record. Fifty-nine such files accumulated there in about two weeks, and the tally across them reads 142 findings at Execution level and 60 at Plan-spec — a clear, actionable reading (the doing is lossy; the specs are mostly sound) that no individual PR could have produced. NetPace ships far more slowly, so the aggregate will take much longer to become legible here. That is an argument for starting the record cheaply now, not for waiting: files never written can never be read back.

Three design points follow from that project's experience:

- **A clean run must write nothing.** With no "nothing notable" entry, the presence of a file means the work taught us something, and the folder stays all-signal. The denominator — how many features shipped without incident — stays recoverable from merged-PR history.
- **A level nobody can populate is worse than no level.** That project also carried a *Goal* level, for work whose aim was itself misaimed. It recorded **zero** in two weeks, because whether an issue should have existed at all is settled when the issue is drafted, not downstream while building it. That judgement is not available to the thing doing the work, so the level sat empty and diluted the tally. It is deliberately absent here.
- **A surprise is not always our process's fault.** Execution and Plan-spec both locate the fault in how we work — the doing, or the drafting. Neither covers a surprise that came from the code as it already was, or from something outside the repository entirely. Those get their own levels, because their fixes have different owners.

## Users & jobs

- **Someone who has just finished a piece of work.** They ran `/build` and `/verify`, something during the run was unexpected, and they want it recorded before the context evaporates — without writing the file by hand or deciding on a format each time.
[...]

PR #275, opened 2026-09-07
Ready Label On Confirm

### LINKED ISSUE #274: Apply the ready label when /speckit.confirmissue folds in decisions
## Summary

`/speckit.confirmissue` finishes by folding the answered review into a `## Confirmed decisions` section, but leaves no mark on the issue's own state. Applying the existing `ready` label at that point makes "this issue is fully defined" visible from the issue list, without opening each one.

## Motivation

The repo already argues for this. [docs/agentic-workflow.md:163](https://github.com/FrankRay78/NetPace/blob/main/docs/agentic-workflow.md#L163): *"Prefer the tracker's own state as that ledger. Before building a progress file, check what the issue tracker already models: issue open/closed, dependency relationships, labels, and PR links are durable, shared, human-visible, and already the thing the team edits… A bespoke ledger has to be kept true; the tracker is true by construction."*

Nothing in `.claude/` writes a label today — no command, skill or hook contains an `--add-label`, `--remove-label` or `--label` write. Every label in this repo is applied by hand, `ready` included (9 issues to date).

Meanwhile the pipeline is about to become label-driven at its *entry*: #271 makes applying `review` the trigger for an automated pre-specification review, and has the workflow remove that label once the review is posted. Nothing marks the *exit*. `ready` already exists, already means *"Issue is fully defined and ready to be implemented"*, and is exactly the state the confirm step produces — so the state machine completes itself: `review` = review pending → removed = reviewed → `ready` = defined.

## Capability

Running the confirm step on an issue whose review is fully answered leaves that issue carrying the `ready` label. Nothing else about the issue's labels changes. If the label cannot be applied, the confirmed decisions are still saved to the issue body and the failure is visible in the command's report.

## Out of scope

- Removing or changing any other label, `needs triage` included — an issue may end up showing both `needs triage` and `ready`, and tidying that stays a manual call.
- Changing the `ready` label's name, colour or description.
- Label handling in any other pipeline command — `/speckit.reviewissue`'s label behaviour belongs to #271.
- Making the label a gate that any downstream command reads. It is a human-visible marker, not a precondition.

## Acceptance criteria

- [ ] An issue that has just been through the confirm step carries the `ready` label.
- [ ] Every label the issue already carried is still on it afterwards — the confirm step adds one label and removes none.
- [ ] Re-running the confirm step on an issue that is already `ready` leaves it `ready` and reports no error.
- [ ] An issue whose review still has an unanswered or hedging gap does not come out labelled `ready`.
- [ ] When the label cannot be applied, the confirmed decisions are still saved to the issue body, and the command's report says the label was not applied.
- [ ] The workflow pipeline description states that the confirm step marks the issue ready.

Evidence is one real run linked in the PR: an issue without ready that carries it after /speckit.confirmissue. Do not stage the failure path — mark AC 5 as design intent, verified by reading the command text rather than by a run. Reason: forcing the label write to fail means inventing a bogus label name purely to watch an error path, which is scaffolding larger than the line it protects, and it matches the "no synthesised tests" call 

## Technical notes

- The change is confined to [.claude/commands/speckit.confirmissue.md](https://github.com/FrankRay78/NetPace/blob/main/.claude/commands/speckit.confirmissue.md) — a label step after the existing body patch, plus a line in the "Output to the user" section — and the pipeline description at [docs/agentic-workflow.md:45](https://github.com/FrankRay78/NetPace/blob/main/docs/agentic-workflow.md#L45), which currently describes the command as only folding the review into a section.
- Apply the label *after* the body patch succeeds, so a label failure can never leave the decisions unsaved. The body patch is the command's real work; the label is a marker on top of it.
- Adding a label an issue already carries is a no-op on GitHub's side, which preserves the command's stated idempotency ("re-running is harmless").
- The existing hard-stops in step 2 (unanswered slot, hedging answer) already guarantee the label can only be applied in a genuinely confirmed state — no new gate is needed for the fourth acceptance criterion.
- Constitution Principle I, configuration/tooling carve-out: this is not production code, so the RED-GREEN evidence is the real command observed leaving an issue unlabelled before the change and labelled after, recorded in the PR.
- **Provenance of the decisions below**: they were settled in a working session rather than through `/speckit.reviewissue`, because a one-line command-file change does not warrant the review cycle. There is no review comment on this issue to audit them against.

### Requirements

- **Label scope:** Add `ready` only — no other label is added, removed or altered, including `needs triage`. An issue showing both `needs triage` and `ready` is accepted as a manual tidy-up rather than something the command resolves.
- **Label wording:** The `ready` label's description stays as-is. "Fully defined" is the operative half and is already accurate; the "ready to be implemented" half is exact for issues that go straight to `/build` and merely loose for those that pass through `/speckit.specify` first.

### Technical

- **Ordering and failure handling:** The label is applied after the issue body has been patched, and a failure to apply it does not fail the command — the decisions are already saved, and the report says the label did not land.
- **Process:** No `/speckit.reviewissue` cycle was run for this issue. The change is a single step in one command file, so a full review-and-confirm pass would have cost more than the change itself.

## Related

- #271 — makes the `review` label the trigger for an automated pre-specification review, and removes it after posting. This issue completes the other end of the same label state machine.
- [docs/agentic-workflow.md:163](https://github.com/FrankRay78/NetPace/blob/main/docs/agentic-workflow.md#L163) — "prefer the tracker's own state as that ledger", the principle this change applies.

FrankRay78:
## Why

`/speckit.confirmissue` finished by patching the issue body and left no mark on the issue's own state, so "this issue is fully defined" was invisible from the issue list — you had to open each one. `ready` already exists and already means exactly that. Applying it here also completes the label state machine #271 starts at the other end: `review` = review pending → removed = reviewed → `ready` = defined.

## What changes

- `/speckit.confirmissue` now applies the `ready` label after it patches the issue body. Additive only — no other label is added, removed or altered, `needs triage` included.
- The command's report says whether the label landed. A label failure is never fatal: the decisions are already saved by then, and the report says so.
- The pipeline description in `docs/agentic-workflow.md` says the confirm step marks the issue ready.

- **The organic evidence run the issue asks for does not exist yet, and I could not produce one.** The issue calls for "one real run linked in the PR: an issue without `ready` that carries it after `/speckit.confirmissue`". No open issue is in a state where the command can complete: only #260 carries a `<!-- speckit:review -->` comment and all ten of its `> _Answer:_` slots are empty, so the command hard-stops at step 2. #274 itself has no review comment and already carries `ready`, applied by hand. Manufacturing a candidate would mean inventing answers on a real issue or opening a throwaway one — scaffolding of the kind this issue explicitly rejects for the failure path. **The next real confirm run on an unlabelled issue is the evidence**; this PR ships without it rather than staging it.
- **This is a configuration/tooling change, not production code**, so per Constitution I's carve-out the RED-GREEN evidence is the real state failing and passing, not an xUnit test. RED: `grep -rn -- "--add-label|--remove-label" .claude/` returned nothing — no command in the repo wrote any label; and #274 carries a `## Confirmed decisions` section yet its `ready` label came from `FrankRay78` in the issue timeline, not from the command. GREEN: the same grep now finds the label write, and line 45 of the workflow doc names it. Full suite run anyway and unaffected: 648 passed, 0 failed, 0 skipped, 0 warnings.
- **AC 5 (label failure keeps the decisions saved) is design intent, verified by reading the command text**, as the issue directed. Staging it would mean inventing a bogus label name purely to watch an error path.
- **No new gate guards AC 4.** An issue with an unanswered or hedging review can't come out `ready` because the label step sits downstream of step 2's existing hard-stops — the ordering is the guarantee, and it is written into the step so a future editor doesn't move it.
- Idempotency (AC 3) rests on GitHub treating `--add-label` of an existing label as a no-op, which preserves the command's stated "re-running is harmless".
- I also added the label to the command's `## Purpose` list, beyond the issue's stated file touchpoints, so that summary doesn't describe a command that now does more than it says.

## How to verify

- [ ] Read the new step 6 in `.claude/commands/speckit.confirmissue.md` — it runs after step 5's body patch, and before step 7's "do not touch the review comment".
- [ ] Confirm nothing in the diff can remove a label: `grep -rn -- "--remove-label" .claude/` finds only the prose forbidding it.
- [ ] Check the ordering claim holds against step 2's hard-stops — an unanswered or hedging answer returns before any body patch, so the label step is unreachable.
- [ ] Next time an issue's review comes back fully answered, run `/speckit.confirmissue #N` on it and confirm the label lands — that is the outstanding evidence noted above.

## Related

Closes #274

- #271 — makes `review` the trigger for the automated pre-specification review and removes it after posting. This PR completes the other end of the same label state machine.
- [docs/agentic-workflow.md:163](https://github.com/FrankRay78/NetPace/blob/main/docs/agentic-workflow.md#L163) — "prefer the tracker's own state as that ledger", the principle this change applies.

### DIFF
--- a/.claude/commands/speckit.confirmissue.md
+++ b/.claude/commands/speckit.confirmissue.md
@@ -27,6 +27,7 @@ Sits **between** `/speckit.reviewissue` and `/speckit.specify`.
 1. Reads the answered review comment.
 2. Pairs each gap's `**Recommendation:**` with the author's `> _Answer:_` to produce a **one-line decision** per gap.
 3. Appends (or rewrites, if already present) a `## Confirmed decisions` section at the end of the issue body.
+4. Applies the `ready` label, marking the issue fully defined without anyone having to open it.
 
 The original review comment is left untouched — it serves as the natural audit trail of how each decision was reached.
 
@@ -158,7 +159,21 @@ gh api --method PATCH repos/<owner>/<repo>/issues/<number> --input .claude/scrat
 (Omit the leading `/` on the endpoint — Git Bash on Windows rewrites `/repos/...`
 as a filesystem path. `gh api` accepts both forms on Linux/macOS.)
 
-### 6. Do not touch the review comment
+### 6. Apply the `ready` label
+
+With the body patched, mark the issue as fully defined:
+
+```bash
+gh issue edit <number> --repo <owner/repo> --add-label ready
+```
+
+- **Additive only.** `--add-label` adds `ready` and touches nothing else — every label the issue already carries stays on it. Never pass `--remove-label` here; tidying `needs triage` or anything else alongside it is a manual call.
+- **Idempotent.** Adding a label an issue already carries is a no-op on GitHub's side, so re-running this command on an already-`ready` issue leaves it `ready` and reports no error.
+- **Never fatal.** If the label cannot be applied, do **not** fail the command — the decisions are already saved. Continue to the report and say there that the label did not land.
+
+Ordering matters both ways: the body patch runs first so a label failure can never leave the decisions unsaved, and this step sits downstream of step 2's hard-stops so an issue with an unanswered or hedging gap can never come out labelled `ready`.
+
+### 7. Do not touch the review comment
 
 The answered review comment is the audit trail. Leave it intact. Do not delete, edit, or annotate it.
 
@@ -170,6 +185,7 @@ Keep your chat response short:
 
 - confirm the issue updated (number + title)
 - state how many decisions were folded in (and the count by pattern: e.g. "8 accepted, 1 with rider, 1 redirected")
+- state whether the `ready` label was applied — and if it was not, say so explicitly, noting the decisions were saved regardless
 - return the issue URL
 
 If you stopped at step 1 or 2, report which precondition failed and what to fix.
--- a/docs/agentic-workflow.md
+++ b/docs/agentic-workflow.md
@@ -42,7 +42,7 @@ A useful frame is the **five duties of a harness** (OpenAI): the harness must **
 ### Per-feature — Spec & planning (on the main branch)
 1.  `/speckit.draftissue`    ← optional; turn an unstructured brief into a well-formed issue
 2.  `/speckit.reviewissue`   ← pre-spec gate; posts gaps + recommendations as an issue comment
-3.  `/speckit.confirmissue`  ← fold answered review into a `## Confirmed decisions` section
+3.  `/speckit.confirmissue`  ← fold answered review into a `## Confirmed decisions` section, and label the issue `ready`
 4.  `/speckit.specify`
 5.  `/speckit.clarify`       ← iterate until the spec feels complete
 6.  `/speckit.checklist`     ← resolve all gaps before continuing


FrankRay78:

- Bugs or correctness issues

PR #276, opened 2026-09-07
Automated Issue Review

### LINKED ISSUE #271: Automated pre-specification review on labelled issues

FrankRay78:
## Why

`/speckit.reviewissue` is the pre-specification gate, and it is the one step in the loop whose cost is the *wait* — it carries the codebase-grounding pass. Running it only from a developer machine means the first step of a three-step loop gates the other two: an issue raised from a phone sits untouched until the author sits down. Labelling an issue `review` now runs it unattended, so the gap analysis is waiting rather than requested.

## What changes

- Applying the `review` label to an issue produces a pre-specification gap analysis as a comment on that issue — numbered gaps, a recommendation per gap, an inline answer slot — grounded in this codebase's real paths and conventions. The author answers inline and runs `/speckit.confirmissue` exactly as before.
- The label is the state. A green run always ends with the label removed, so **labelled** = review pending, **unlabelled + a `<!-- speckit:review -->` comment** = done, **still labelled** = the run did not complete. There is no failure comment; GitHub's failed-run notification is the alert.
- Labelling an issue that already carries a review posts nothing and clears the label, rather than posting a second review that would orphan answers already written against the first.
- `claude.yml` is untouched, so `@claude` on issues and PRs behaves exactly as before.

The review logic itself is not duplicated — the workflow prompt reads [`.claude/commands/speckit.reviewissue.md`](.claude/commands/speckit.reviewissue.md) and applies it, so the analysis stays single-sourced and a change to the command file takes effect without touching the workflow.

**Three deliberate deviations from the issue's Confirmed decisions**, each because the decision as written was not implementable or did not hold:

1. **The workflow snapshots the issue with `gh` instead of reading the context the action injects.** The confirmed decision said injection made fetching unnecessary. It does not: the action *sanitises* injected content by stripping HTML comments, which would destroy the `<!-- speckit:review -->` sentinel that the guard, `/speckit.confirmissue`, and the entire design key on. Separately, `workflow_dispatch` carries no issue payload to inject at all.
2. **The `if:` gained a `workflow_dispatch` disjunct.** The confirmed expression alone would have made the dispatch path — which the same decision mandates — unreachable.
3. **`--allowedTools` gained `Read,Glob,Grep,Write`.** The prompt requires the Write tool, and the whole codebase-grounding pass requires reads. The confirmed decision was about which *writes to GitHub* Claude gets; both `Bash(gh …)` grants are unchanged.

**Two defects found in review that broke the contract the confirmed decisions describe**, both fixed in `b1d6d00`:

- `claude-code-action` exits green whenever the model finishes its turn — a failed `gh issue comment` is tool output the model reads, not a step failure. A prompt instruction saying "let the job fail" cannot be honoured. So the run that posted nothing went green and notified nobody, which is precisely the state "GitHub's failed-run notification is the alert" was meant to cover. There is now a verification step that re-reads the issue and fails the job unless the sentinel is present *and* the label is gone.
- The already-reviewed guard originally left the label attached, producing *labelled + review comment* — a state the doc's own model reads as "pending", permanently, and reachable by the ordinary gesture of reviewing locally then labelling. The guard is now a `run:` step (a model can skip a prompt instruction) and clears the label.

**Known residuals**, all recorded in the CIR:

- **The tool allowlist is not a boundary.** `--allowedTools` *adds* to the permission rules Claude Code loads from the checked-out `.claude/settings.json`, which already grants `Edit(**)`, `Bash(gh:*)` and `defaultMode: acceptEdits`. The real grant is every `gh` subcommand, and the prompt's "do not edit the issue body" is a request, not an enforced limit. Narrowing it properly means changing settings that govern every local session — separate work.
- **The snapshot bypasses the action's sanitiser.** Keeping the sentinel intact necessarily means the raw, attacker-authorable issue body reaches the model unsanitised while it holds an `issues: write` token. The label gate is the mitigation: untrusted text only reaches the model if a maintainer deliberately labels it.
- **A label applied by any account other than `FrankRay78` is silent.** The job-level `if:` means GitHub creates no run, so nothing appears in the Actions tab and no notification fires — yet the resulting state (labelled, no review) is the one reserved for a failed run. The gate expression is pinned by the confirmed decisions, so this is flagged rather than changed.

`fetch-depth: 0` and `id-token: write` are both unused here and were flagged in review. Kept deliberately for parity with the known-working `claude.yml`: this workflow cannot be triggered before merge, so changes were confined to correctness rather than untested tidying.

## RED–GREEN evidence (Constitution Principle I, configuration/tooling carve-out)

The gate is the real trigger, not a hand-rolled test. No bespoke test was written to stand in for GitHub Actions.

**RED**, before the change — parsing every workflow's `on:` block:

```
Workflows triggered by the issues event: NONE
$ gh run list --limit 50 --jq '[.[]|select(.event=="issues")]|length'
0
```

**GREEN**, structural, after:

```
.github/workflows/speckit-reviewissue.yml: on = ['issues', 'workflow_dispatch']
issues trigger types: ['labeled']
```

**GREEN, end-to-end, is merge-then-verify and is still outstanding** — no event can trigger a *new* workflow file from a feature branch. Per the issue's own confirmed decision, the completing evidence is the real label trigger observed working on `main` against a throwaway issue. I will link that run here after merge.

## How to verify

- [ ] `.claude/commands/speckit.reviewissue.md` is unmodified — the analysis must stay single-sourced, and the prompt delegates rather than restating it.
- [ ] The prompt's three declared deltas match the command file: step 1 (fetch/mode detection, but *not* its "don't re-raise a resolved gap" rule, which is explicitly retained), step 4's `gh repo view` default-branch resolution, step 5's heredoc.
- [ ] `claude.yml` is untouched; mention `@claude` on any issue or PR and confirm unchanged behaviour.
- [ ] **After merge:** label a throwaway issue `review`. Confirm a review comment appears citing at least three repo-specific paths, and that the `review` label is removed.
- [ ] **After merge:** re-apply `review` to that same issue. Confirm no second comment is posted and the label is cleared again.
- [ ] **After merge:** confirm the run that produced the review is green, and that `workflow_dispatch` with an `issue_number` input works against a branch ref.

## Related

Closes #271

- [`.claude/commands/speckit.reviewissue.md`](.claude/commands/speckit.reviewissue.md) — the review logic this automates.
- [`.claude/commands/speckit.confirmissue.md`](.claude/commands/speckit.confirmissue.md) — the downstream step that consumes the review comment; it finds the review by sentinel and does not filter on author, which is what lets a CI-posted comment be folded in unchanged.
- [`docs/change-intent-records/2026-09-07-automated-prespec-review.md`](docs/change-intent-records/2026-09-07-automated-prespec-review.md) — decisions and residuals.


FrankRay78:

- Bugs or correctness issues

PR #283, opened 2026-09-11
Delete Review Comment

FrankRay78:
## Why

The pre-specification review comment is working material. A numbered gap carries framing, cited evidence, a recommendation with its reason, and an answer slot — and ten of those compress to ten one-line decisions once `/speckit.confirmissue` has folded them in. After that the comment dwarfs the `## Confirmed decisions` section that matters, and whoever picks the issue up has to work out which of the two halves is current. In practice it was already being deleted after every successful run; the prompt forbade the thing that actually happened.

Deleting it is the easy half. The comment's *existence* was the done-marker — `docs/agentic-workflow-NetPace.md` stated the contract, and two guards keyed on it — so the marker had to move first, or "confirmed" would have become indistinguishable from "never reviewed".

## What changes

**The marker moves to the `ready` label.** `/speckit.reviewissue` and the `speckit-reviewissue.yml` guard both check it before falling back to the `<!-- speckit:review -->` sentinel, so issues confirmed under the old behaviour still read as reviewed with no migration.

**`/speckit.confirmissue` deletes the review**, but only after reading the `ready` label and the `## Confirmed decisions` section back from the issue. Every hard-stop sits upstream of that point, so a run that stops early leaves the review answerable and re-runnable.

**Re-running confirmation on a confirmed issue now says so**, instead of reporting that there is no review to confirm — and repairs a missing `ready` label while it is there.

**Reopening is removing the `ready` label.** An issue whose scope has moved is by definition no longer ready; nobody has to hand-edit the body.

**Docs:** the state contract in `docs/agentic-workflow-NetPace.md`, a new CIR, and in-place supersede annotations on `2026-09-07-automated-prespec-review.md`.

## Non-obvious things a reviewer should know

**The marker is the label alone, not "label OR body section".** The issue's technical notes sketched the OR, but it cannot coexist with AC #8 — the only way to clear a body section is to hand-edit the body, which is exactly what AC #8 forbids of the reopen path. CIR decision 1 records this and both rejected alternatives. This is the one place the implementation deliberately departs from the issue's own technical sketch.

**Deletion is gated on the label having landed, not merely on the body patch.** The label step is never-fatal by design, so deleting after a label failure would leave decisions on the body with no label and no comment — indistinguishable from never-reviewed to both guards, and re-reviewed on top of its own decisions. That state is the one thing the design exists to prevent, and it is the only one reached through a door that cannot be reopened.

**Three blockers were found in review and are fixed here** — worth knowing because the first would have made the feature a silent no-op:

[...]

PR #282, opened 2026-09-11
Commentary Section

FrankRay78:
## Why

`/speckit.reviewissue` emitted a section called **Notes for SDD** — a name that describes a pipeline stage rather than its content, and that only resolves for a reader who already knows the answer. A "Speckit Docs Tidy" commit ([e2996b0](https://github.com/FrankRay78/NetPace/commit/e2996b0)) had also given it a second flavour: items "deferred to `/speckit.specify`". That hand-off never worked and never could — `/speckit.specify` is stock spec-kit taking a natural-language description via `$ARGUMENTS`, with no notion of a GitHub issue, a comment or a review. The same commit stated the deferral rule twice, in two different wordings, two paragraphs apart.

The section also had no written boundary and no bar for entry, so it had already absorbed content outside its own definition once and was carrying bullets of the right kind and no value ("three hooks will fire, all of them no-ops here").

## What changes

- The section is **Commentary**, everywhere it appears including the emitted comment template.
- The `/speckit.specify` hand-off is gone, along with the duplicate routing rule that expressed the same drift without naming a destination.
- The section now states what it carries: remarks that inform the spec author and require no answer, and nothing else.
- **A boundary**: the source of a constraint decides, not its force. A fact already true of the codebase is commentary however binding it proves; a choice only the author can make, or an obligation this issue would newly impose, is a numbered gap.
- **A filter**: only what the spec author would otherwise miss or get wrong. A bullet restating a rule they already hold, or amounting to "nothing to do here", does not appear — and the template now omits the heading entirely rather than emitting it empty.
- The example list was checked against the new filter and the two examples it bans were removed.

## RED–GREEN evidence

Constitution Principle I's configuration/tooling carve-out applies: this is a prompt, so the evidence is the real check failing and then passing, not a bespoke test. No hand-rolled checker was written.

**RED** — before the change:

```
$ grep -n "Notes for SDD" .claude/commands/speckit.reviewissue.md
74:  ... plus a **Notes for SDD** section for non-question observations.
99:  ... Items the author wants tracked but not answered now belong in **Notes for SDD**.
109: **Notes for SDD** — ... items the author chose to defer to `/speckit.specify` ...
166: ### Notes for SDD
241: ... The "Notes for SDD" section is also untouched.

$ grep -nEi "\bnotes?\b" .claude/commands/speckit.reviewissue.md
277: - state how many gaps + how many notes were raised
```

**GREEN** — after:

```
[...]

Issue #280, opened 2026-09-11
Delete the review comment once /speckit.confirmissue has folded it into the issue body


FrankRay78:
## Summary

`/speckit.confirmissue` deletes the review comment once it has folded the answers into the issue body, and the "this issue has been reviewed" signal moves off the comment's existence and onto the `ready` label and the `## Confirmed decisions` section. Someone picking up a confirmed issue sees the decisions, not the long gap analysis that produced them.

## Motivation

**The review comment is the wrong thing to hand whoever picks the issue up.** It is large by design — a numbered gap carries framing, cited evidence, a recommendation with its reason, and an answer slot. A review of ten gaps compresses to ten one-line decisions in the issue body. Once that compression has happened the comment is working material: it dwarfs the section that matters, and a reader arriving at a confirmed issue has to work out which of the two halves is current. In practice it is already deleted after every successful run for exactly that reason.

**The prompt forbids the thing that actually happens.** [speckit.confirmissue.md:32](.claude/commands/speckit.confirmissue.md) says *"The original review comment is left untouched — it serves as the natural audit trail of how each decision was reached"*, and step 7 at [speckit.confirmissue.md:176-178](.claude/commands/speckit.confirmissue.md) says *"Leave it intact. Do not delete, edit, or annotate it."* The rule is wrong, not the practice. The confirmed-decision bullets already carry each gap's title, its resolved outcome, and `(Author redirected from "…")` where the author overruled the recommendation — which is the causality anyone needs afterwards. The gap framing is how the decision was reached in the moment, not a record worth keeping.

**The comment's existence is load-bearing elsewhere, and that is the real work.** [docs/agentic-workflow-NetPace.md:34](docs/agentic-workflow-NetPace.md) states the contract: *"labelled means a review is pending, **unlabelled with a `<!-- speckit:review -->` comment** means it is done, and still labelled means the run did not complete."* Delete the comment and "done" becomes indistinguishable from "never reviewed". Two guards key on it:

- [speckit-reviewissue.yml:76-86](.github/workflows/speckit-reviewissue.yml) greps the comment list for the sentinel and skips the run if it finds one. With the comment gone, re-applying the `review` label to a confirmed issue passes the guard and posts a fresh full review on top of `## Confirmed decisions`, re-asking questions that were answered and folded weeks earlier.
[...]

Issue #279, opened 2026-09-11
Rename Notes for SDD to Commentary, and give the section a boundary and a filter


FrankRay78:
## Summary

Four surgical changes to `/speckit.reviewissue`, all to one section: rename **Notes for SDD** to **Commentary**, delete a downstream hand-off that a docs-tidy commit invented and that cannot work, state plainly what the section may and may not carry, and filter what reaches it on non-obviousness. The section stops naming a pipeline stage it has no connection to, stops promising a consumer that never reads it, and gains both the boundary and the filter it needs before anything widens its intake.

## Motivation

**The name describes a pipeline stage, not its content.** "Notes for SDD" tells a reader nothing about what belongs in the section — the acronym only resolves if you already know the answer, and it implies the content is addressed to a later spec-driven-development step. "Commentary" says what it actually is: remarks that inform the reader without asking anything of them.

**The downstream hand-off was never designed — it drifted in.** The section's first version ([5ffd34d](https://github.com/FrankRay78/NetPace/commit/5ffd34d)) had one flavour and no downstream consumer:

> **Notes for SDD** — things the spec author should *know* but don't need to answer. These are observations that will shape the spec without being open questions

A later commit titled **"Speckit Docs Tidy"** ([e2996b0](https://github.com/FrankRay78/NetPace/commit/e2996b0)) added a second flavour:

> Two flavours land in this section: passive observations..., and items the author chose to defer to `/speckit.specify` rather than resolve at issue level

That hand-off cannot work, and never could:

- `/speckit.specify` is stock spec-kit, not a NetPace command — `.claude/skills/speckit-specify/SKILL.md`, `metadata.source: templates/commands/specify.md`. Its input is a natural-language feature description via `$ARGUMENTS`. It contains no reference to a GitHub issue, a comment, or a review.
- `/speckit.confirmissue` never reads the section either. It folds numbered gaps into the issue body and has no mention of notes at all.
- The review comment does not survive to be read by any later stage in practice — see *Open questions* below.

The same commit put the deferral flavour in the prompt twice, in two different words. Line 109 names `/speckit.specify` explicitly; line 99 says the same thing without naming a destination — *"Items the author wants tracked but not answered now belong in **Notes for SDD**"*. Both are the same drift, and both have to go: renaming line 99 while leaving its routing rule intact would leave the prompt contradicting the boundary added two paragraphs below it.

[...]

PR #284, opened 2026-09-12
Triageable Review

### LINKED ISSUE #281: Make a /speckit.reviewissue comment triageable, and stop the category sweep inflating it
## Summary

A `/speckit.reviewissue` comment becomes triageable: the author sees every gap's number, title and consequence before reading any of them, the gaps that change what gets built come first within their group, each gap is bounded in length, and the review raises gaps where the issue leaves something open rather than one per category it was prompted to consider. The yield is unchanged — no gap is dropped, demoted or capped on merit. What changes is that attention can be spent in proportion to what is at stake.

## Motivation

**The output is read-in-full or skipped, with no middle.** Every gap arrives at identical visual weight: the same bold numbered heading, the same framing paragraph, the same recommendation, the same answer slot. Nothing distinguishes a gap that changes what gets built from one that settles a file name. The only way to find out which is which is to read all of it, and the only alternative to reading all of it is to read none of it. That is the choice the comment currently offers, and skipping is the rational response to it when the review is long.

Four properties of the prompt produce that, all of them checkable at HEAD:

- **No summary.** The template opens straight into the instruction block and the first numbered gap ([speckit.reviewissue.md:124-137](.claude/commands/speckit.reviewissue.md)). There is no point at which the author sees the shape of what they are being asked before they are already inside the first question.
- **No length bound.** The nearest thing is *"Keep each gap tight. If a gap has more than ~3 sub-bullets, consider whether it is actually two gaps"* ([speckit.reviewissue.md:172-173](.claude/commands/speckit.reviewissue.md)). That bounds sub-bullets, not prose, and its remedy — split the gap — raises the gap count while leaving the word count where it was.
- **No ordering by consequence.** Gaps are numbered contiguously across two groups ([speckit.reviewissue.md:87](.claude/commands/speckit.reviewissue.md)) which are Requirements-then-Technical ([:89](.claude/commands/speckit.reviewissue.md), [:99](.claude/commands/speckit.reviewissue.md)). Position therefore encodes category and nothing else. A gap that would change an acceptance criterion can sit below one that settles a name, and the reader has no way to know until they have read both.
- **A mandated sweep.** *"Cover at least these categories when applicable"* ([:89](.claude/commands/speckit.reviewissue.md)) is followed by seven requirements categories ([:91-97](.claude/commands/speckit.reviewissue.md)) and seven technical ones ([:101-107](.claude/commands/speckit.reviewissue.md)). Fourteen prompted categories, each a standing invitation to find something. Breadth is instructed; proportion to the issue in hand is not.

**What must not change is the yield.** Gaps that change what gets built are the product of this command, and a review that surfaces several of them has done its job even when it is long. So nothing here reduces their number: there is no test that demotes a point on a prediction about whether the author would contest it, no cap, and no route that sends a decision somewhere it carries no answer slot. This issue changes what the author sees first, how long a gap may run, the order gaps arrive in, and how wide the net is cast before any of that.

## Users & jobs

- **The issue author opening a review.** They want to spend their attention in proportion to what is at stake — settle the two or three gaps that change the build, move quickly through the ones that settle a detail — without reading everything to discover which is which.
- **The same author returning to a half-answered review.** They want to see what is still outstanding, and how much of it matters, without scrolling the whole comment again.
- **The author of a small, tightly-scoped issue.** They want a review proportionate to the issue. An issue that commits to little should not attract a review sized by the number of categories the command was told to consider.

## Capability

**Scenario: A review is triaged before it is read**
Given an issue whose review raises several gaps, When the author opens the comment, Then they can see every gap's number, title and consequence without reading into the gaps themselves.

**Scenario: The gaps that change the build come first**
Given a review whose gaps differ in consequence — some would change what the issue commits to building, others settle a detail within it — When the comment is composed, Then within each gap group the higher-consequence gaps come first, and each gap states which kind it is.

**Scenario: A small issue gets a small review**
Given an issue that leaves little open, When the review runs, Then it raises gaps only where something is genuinely unsettled, and the review is visibly shorter than one of an issue that leaves much open.

## Out of scope

- Deciding whether a point becomes a gap by predicting whether the author would contest the recommendation. That judgement is about a person rather than about the issue in front of it, and a wrong call settles a real decision silently. Whether the issue leaves something genuinely open *is* in scope — narrowing the sweep on that basis is one of this issue's four levers — but every gap admitted that way keeps its answer slot, and nothing is demoted into a section that has none.
- Routing anything into Commentary. #279 settles that Commentary carries only remarks requiring no answer, and that a choice the author must make, or an obligation the issue would newly impose, is a numbered gap. This issue adds nothing to Commentary's intake and changes none of its rules.
- A numeric cap on the number of gaps. The sweep is narrowed at source instead; a cap layered on top would force out a genuine gap for arriving last.
- Changing the mandatory `**Recommendation:**` rule, the `> _Answer:_` slot format, the existence or order of the two gap groups, the refine-run rules, or the link rules.
- Changing what `/speckit.confirmissue` folds into the issue body, or anything in its prompt.
- The review comment's lifetime and the state marker that replaces it — #280.
- Migrating review comments already posted in the old shape. Existing comments stay as they are, and refine runs against them continue to work.
- Changes to `/speckit.draftissue`, `/speckit.testplan`, `/speckit.testchecklist`, or any stock `speckit-*` skill.

## Acceptance criteria

- [ ] A reader can see the whole set of gaps — number, title and consequence — before reading any gap body.
- [ ] Every gap states what changes if it is answered against its recommendation: something the issue commits to building, or a detail within it.
- [ ] Within each gap group, gaps that change what the issue commits to building come before those that settle a detail, and the numbering follows that order.
- [ ] The Requirements group still precedes the Technical group, and numbering is still contiguous across both.
- [ ] No gap exceeds the stated length bound, and the bound applies to the gap as a whole rather than to its sub-bullets.
- [ ] The review raises a gap only where the issue leaves something open — a category with nothing unsettled in it produces no gap, rather than a gap manufactured to cover the category.
- [ ] A review of an issue that leaves little open is visibly shorter than a review of one that leaves much open.
- [ ] Every gap still carries a recommendation with a one-sentence reason and an answer slot, unchanged in form.
- [ ] `/speckit.confirmissue` folds exactly the numbered gaps, with no change to its own prompt — nothing added above the gaps is parsed as a gap.
- [ ] After a refine run, what the reader sees up front still matches the gaps it describes.
- [ ] A refine run leaves every non-hedging gap byte-for-byte unchanged, as it does today.

---

### Where it lives

`.claude/commands/speckit.reviewissue.md` only. #279 and #280 both edit this file first, so the regions below are named by what they do; line numbers are given only where neither of those issues moves anything.

- **The category lists in *Identify gaps*** ([:89](.claude/commands/speckit.reviewissue.md), [:91-97](.claude/commands/speckit.reviewissue.md), [:99](.claude/commands/speckit.reviewissue.md), [:101-107](.claude/commands/speckit.reviewissue.md)) — the sweep narrows here. The categories stay as a checklist of what to *consider*; what changes is that considering one and finding nothing open must produce nothing.
- **The numbering rule** ([:87](.claude/commands/speckit.reviewissue.md)) — contiguous numbering is retained and gains an ordering rule within each group.
- **The comment template** ([:124-170](.claude/commands/speckit.reviewissue.md)) — gains the up-front view of the gaps, above the first group.
- **The gap-quality rules** ([:172-181](.claude/commands/speckit.reviewissue.md)) — the length bound lands alongside *"Keep each gap tight"*, which it replaces as the operative constraint.
- **Step 6's refine rules** — re-framing a hedging gap must keep the up-front view in step with it.
- **The recommendation rule** ([:80-85](.claude/commands/speckit.reviewissue.md)) is unchanged. It constrains gaps that exist and is not what decides that one exists.

### Constraints / assumptions

- `/speckit.confirmissue` parses each `**N. <title>**` block ending in a `> _Answer:_` line. Whatever form the up-front view takes, it must not produce text matching that shape, or it will be folded into the issue body as a phantom decision. This is the binding constraint on its design.
- The two gap groups and their order are load-bearing downstream: `/speckit.confirmissue` routes bullets to `### Requirements` or `### Technical` by which group the gap sat under. Consequence ordering therefore operates *within* a group, never across the two.
- The author refers to gaps by number in chat, and `/speckit.confirmissue` strips the numbers when folding, so ordering must be settled before numbering and must not renumber a posted review.
- Prompt-only change. Per Constitution Principle I's tooling carve-out there is no xUnit test and no bespoke checker; the RED-GREEN evidence is the command's own output on a real issue before and after, recorded in the PR body.
- Depends on #280, and transitively on #279. All three edit this prompt, and #280 also changes the comment's lifetime and the command's mode detection; landing last avoids conflicting edits and lets this issue be written against the final shape.

## Open questions / future work

- **Whether the consequence judgement is well-calibrated.** It asks the model to assess, from the issue itself, which of the issue's commitments a different answer would change — a question about the artefact, not a prediction about the author, which is why it is the safer of the two. It is still a model judgement. If it proves unreliable in use, the fallback is ordering by consequence without publishing the marker, which keeps the benefit that survives a wrong call and drops the one that does not.
- **Whether the recommendations themselves hold up.** The value of a short review rests on each recommendation being a good default and its *Reason* citing evidence that survives being checked against the file it names. Nothing in the command verifies either today, and nothing here adds it. That is a separate issue, and a more consequential one than length.
- **Whether a long review is ever the right output.** This issue assumes a large, genuinely open issue should still produce a large review, and only bounds the per-gap prose. If that turns out to be wrong the answer is a conversation about issue size, not a cap on the review.

## Related

- **Supersedes #278**, which diagnosed the same complaint as a problem of admitting too many gaps and proposed demoting the ones the reviewer expected not to be contested. That remedy asked the model to predict which of its own recommendations the author would overrule, which is not a judgement it can be relied on to make, and the cost of a wrong call was a decision settled silently. This issue treats the same complaint as a problem of presentation and proportion, and settles nothing on the author's behalf.
- **#280** — must land first; it changes this prompt's mode detection and the comment's lifetime.
- **#279** — settles what Commentary may carry. Nothing here is routed into it.
- [.claude/commands/speckit.reviewissue.md](.claude/commands/speckit.reviewissue.md) — the only file this issue changes.
- [.claude/commands/speckit.confirmissue.md](.claude/commands/speckit.confirmissue.md) — the downstream consumer whose parser and group routing constrain what may be added to the comment.

FrankRay78:
## Why

The review comment offered two options: read all of it, or read none of it. Every gap arrived at identical visual weight — same bold numbered heading, same framing, same recommendation, same answer slot — so nothing distinguished a gap that would change an acceptance criterion from one that settled a file name. Finding out which was which meant reading both. Skipping is the rational response to that when the review is long.

Four properties of the prompt produced it: no summary, no length bound, no ordering by consequence, and a mandated sweep of fourteen categories where breadth was instructed but proportion to the issue in hand was not.

## What changes

- **An at-a-glance table** above the first group gives every gap's number, title and consequence, so the review can be triaged before it is read.
- **Every gap names its kind** — *Changes what gets built* or *Settles a detail* — and says in one clause what a different answer would change.
- **Gaps are ordered by consequence within each group**, then numbered. Requirements still precedes Technical.
- **A 120-word bound per gap** replaces "keep each gap tight", which bounded *sub-bullets* and offered splitting as the remedy — raising the gap count while leaving the word count where it was.
- **The category lists become things to consider, not a sweep to complete.** A category with nothing open produces nothing: no gap, no placeholder, no commentary bullet saying it is fine.

Yield is unchanged, which is the constraint the issue was most insistent about. No gap is dropped, demoted or capped on merit, and every gap keeps its recommendation and answer slot in exactly the form `/speckit.confirmissue` already parses.

**A downstream parser chose the table's shape.** The obvious index mirrors the gaps — a bold-numbered list — and that is precisely what `/speckit.confirmissue` parses as a gap. #281 puts that command's prompt out of scope, so the constraint had to be designed around rather than fixed. A table row carries a bare number and plain text and matches nothing the parser looks for.

**The failure mode that rules this out is not the obvious one.** The first draft of this branch said a table row would be "folded as a phantom decision the author never made". It would not: `/speckit.confirmissue` hard-stops in step 2 whenever a parsed gap looks unanswered, so a row in gap shape either blocks the whole fold, or its block-end resolves to the *first real gap's* answer slot and corrupts the decisions that do land. Corrected here, and it is a stronger argument for the rule than the phantom was.

**What is load-bearing is the grouping, not the section order.** The first draft claimed "this order is load-bearing". It is not — that command routes each folded decision by the *heading* its gap sat under, so swapping the two sections leaves its output byte-identical. Requirements-before-Technical is retained on its own merits (requirements are probed first; #281 puts the groups' order out of scope), not because the parser needs it. This matters because labelling a false claim "load-bearing" invites a future editor to test it, find it false, and relax the genuine constraint next to it.

**The bound is scoped to first composition, deliberately.** Step 6 requires a re-framed hedging gap to *gain* 2–4 options with trade-offs plus a worked example. That cannot fit in 120 words, and an unscoped bound would have forced a refine run to silently break one of two instructions.

**A pre-post self-check was added because composing to a bound is not meeting one.** The first draft written under this prompt overshot on 3 of 3 gaps, unaided, and no instruction caught it. The check is self-administered rather than gated — see the CIR's residual for why a hook cannot reach this invariant: a first run pipes the comment body straight into `gh issue comment`, leaving no committed artefact to check.

**The zero-gap case is deliberately left alone.** An earlier draft added a "no gaps at all → omit the table" exception. The template has conditional-omit directives for Technical gaps and Commentary but none for Requirements gaps, and the intro sentence promises points that would not exist — so the exception had no support and half-fixing it was worse than not touching it. That path was gap 3 of #278's own review and belongs with #278's successor.

**The taxonomy does not cover process artefacts.** A gap asking which release a change ships under, or what its RED step is, is about neither built behaviour nor a detail within it. The "when borderline, take *Settles a detail*" tie-break absorbed both cases in trial without a wrong call, but the two kinds are not a complete partition. Recorded as a residual.

## How to verify

Per Constitution Principle I's configuration-and-tooling carve-out there is no xUnit test here — the evidence is the command's own output before and after. **The controlled half is done; two live checks are not, and should happen before merge.**

A blind A/B was run to avoid grading my own work: two clean-context agents each composed a review of the **same** issue (#242), one given the pre-change prompt and one the new prompt, neither told a comparison was happening or what was being tested.

| | pre-change | this branch |
|---|---|---|
| Gaps raised | 6 | 5 |
| Words per gap | 171–240 (mean 205) | 103–117 (mean 112) |
| **Over the 120-word bound** | **6 of 6** | **0 of 5** |
| Up-front table | absent | present, titles verbatim |
| Consequence lines | 0 | 5 |

- [x] **AC5** — every gap breached the bound under the old prompt and none under the new one, unaided. That rule exists only on this branch.
- [x] **AC1, AC2** — table with number/title/consequence; every gap carries a consequence line, and rows agree with gap lines exactly.
- [x] **AC4, AC8** — contiguous 1–5, Requirements first; 5 recommendations, 5 reasons, 5 answer slots.
- [x] **AC9 (mechanically)** — zero bold-numbered forms and zero `> _Answer:_` lines above the first group, checked by script against the composed body.
- [x] `dotnet build ./src` clean (0 warnings), `dotnet test ./src` green — 648 passed, 0 failed, 0 skipped. No C# changed; this is the unconditional gate, not evidence for the change.
- [ ] **AC9 live** — post a review on a throwaway issue, answer it, and run `/speckit.confirmissue` end to end. Confirm the table contributes no bullet to `## Confirmed decisions`. The mechanical check above is inspection, not execution.
- [ ] **AC10, AC11 live** — answer one gap with a hedge, re-run `/speckit.reviewissue`, and confirm the re-framed gap's consequence line *and* its table row both move, every other gap and row comes through byte-for-byte, and nothing is renumbered.
- [ ] **AC7** — run the new prompt against a large, genuinely open issue and confirm its review is visibly longer than #242's. Not evidenced: one A/B pair cannot separate prompt effect from agent variation.

Two caveats a reviewer should weigh rather than take on trust. **AC3's ordering rule was only trivially exercised** — all three Requirements gaps came out *Changes what gets built* and both Technical ones *Settles a detail*, so nothing actually needed sorting within a group. And the **pre-change run produced more informative gap titles**: declarative findings ("The CSV writer evaluates the placeholder before discarding it, so the obvious null conversion breaks the build") against this branch's questions ("How do the writers hold a result that may not exist?"). It also found two things the new run missed. I cannot attribute that to the prompt — that agent compiled the change to find it and the other did not — but it is the one signal pointing against this change, and it deserves a human eye.

## Related

Closes #281

- #279 (PR #282) and #280 (PR #283) — both edit this prompt and had to land first; this was written against their final shape.
- **Supersedes #278**, which diagnosed the same complaint as admitting too many gaps and proposed demoting the ones the reviewer expected not to be contested. This branch writes that approach out explicitly: it asks the model to judge a person rather than the issue, and a wrong call settles a real decision silently.
- `docs/change-intent-records/2026-09-12-triageable-review-comment.md` — six decisions, five residuals, including the un-gated enforcement and the `/speckit.draftissue` third consumer that reviews of the two preceding issues both missed.
- One correction worth flagging in the record: commit `c5f4eed`'s message cites the #278 gaps as "184 to 317 words". The 317 was a counting script running past the answer slot to end-of-file, adding 94 words of trailing heading and commentary. The real spread is 184–223; the CIR is corrected, but that commit message is immutable.


FrankRay78:

- Bugs or correctness issues

PR #288, opened 2026-09-12
Table Pipe Escaping

### LINKED ISSUE #286: Escaping rule for a literal pipe in an at-a-glance table cell
## Problem

The at-a-glance table added to `/speckit.reviewissue` by #281 is a Markdown table, and its cells are delimited by `|`. Nothing in the template or its rules says what happens when a gap's short title, or its compressed consequence text, contains a literal `|`.

A title referencing a bitwise or, a pipe-delimited format, a shell pipeline, or a C# pattern like `is A or B` written with `|` would break the row it sits in — silently, because a Markdown table with a stray delimiter renders as a misaligned table rather than as an error.

## Motivation

What makes this worth recording rather than just fixing is *how* it was missed. The table's shape was scrutinised at length against the one thing that consumes it: `/speckit.confirmissue` parses every `**N. <title>**` block ending in `> _Answer:_` and folds it into the issue body, so the obvious bold-numbered index would have been folded as a gap that does not exist. That analysis is in the #281 CIR as decision 2, and it drove the choice of a table over a list.

Having hardened the structure against its *consumer's* parser, nobody checked it against its *host format's* own syntax — a simpler and more likely break than the one that got all the attention. The `@claude` review on #284 found it after merge:

> the table is Markdown, and cells are delimited by `|`. Neither the template nor the surrounding rules say what to do if a gap's short title or compressed consequence text itself contains a literal `|` … worth a one-line escaping note (`\|`) if it's ever hit — not a blocker.

It is already self-demonstrating: `docs/study/281.md`'s row about this finding is the only row in that file that needed `\|` escapes to author.

## Proposal

Add an escaping rule to the table's rules in `.claude/commands/speckit.reviewissue.md`: a `|` appearing in a gap title or consequence cell must be written `\|`.

Worth deciding at the same time whether the rule belongs on the gap titles themselves rather than only on the cells, since the title is copied verbatim into the row — which means the escaping decision is really about what a gap title is allowed to contain.

## Acceptance criteria (draft — expect the review to sharpen these)

- [ ] A gap whose title contains a literal `|` produces a table row that renders as a correct three-column row.
- [ ] The rule states where the escape applies — the title cell, the consequence cell, or the gap title at source.
- [ ] The verbatim-title rule and the escaping rule do not contradict each other: a title escaped in the row still matches its gap unambiguously.
- [ ] The pre-post self-check added by #281 covers this, so a broken row is caught before the comment is posted rather than after.

## Out of scope

- Any change to `/speckit.confirmissue`, whose parser is unaffected: a `\|` in a cell does not create or destroy a `**N. <title>**` block.
- Revisiting the table-versus-list decision. The parser analysis behind it stands; this is a gap in it, not a refutation of it.
- Escaping anywhere else in the review comment. Gap bodies, recommendations and commentary are ordinary prose where `|` is harmless.
- Migrating review comments already posted.

### Requirements

- **Where the escape applies:** Every `|` in any content inside the at-a-glance table is written `\|`; gap titles stay free to contain `|`, and a title cell differing only by that escape still counts as verbatim. Nothing outside the table is escaped or changed.
- **Pipes inside inline code:** The rule states explicitly that it applies inside inline code spans too, since GitHub splits table columns on an unescaped pipe before rendering code spans.

### Technical

- **Pre-post self-check:** Add a clause to "Check the draft before posting": no unescaped pipe anywhere in the table, including inside inline code, so a broken row is caught before posting. It stays a self-check, not a CI gate; nothing outside the table in the review comment is escaped.

## Related

- #281 / PR #284 — added the table; its CIR decision 2 records the parser analysis that this slipped past.
- PR #284's `@claude` review — where this was found, after merge.
- `docs/study/281.md` — the last row records this finding, and needs `\|` escapes to state it.
- `.claude/commands/speckit.confirmissue.md` — the downstream consumer the table *was* hardened against.

FrankRay78:
## Why

The at-a-glance table that #281 added to `/speckit.reviewissue` is a Markdown table. A literal `|` in a gap title or consequence cell silently splits the row into the wrong columns. It shows as a misaligned table, not an error. The `@claude` review on #284 caught this after merge.

## What changes

- Every `|` in a table cell's content is written `\|`, including inside inline code. Gap titles may still contain `|`, and a title cell that differs only by the escape still counts as verbatim. Nothing outside the table is escaped.
- The pre-post self-check gains a clause for unescaped pipes in the table.

- **Inline code needs the escape too.** GitHub splits a row into columns *before* it renders code spans. Rendered via `gh api markdown` (gfm): `` `A | B` `` in a cell lost the third column, while `` `A \| B` `` renders as `A | B`.
- **Re-runs check only the row they change.** When a re-run edits an existing review, the pipe check covers just that row. The re-run rules require every other row to pass through byte-for-byte, and migrating already-posted comments is out of scope. This came from the `/verify` review.
- **No CI gate.** The rule governs how a review comment is composed, which nothing in the repo runs, so it stays a self-check, as the issue's confirmed decisions say.

## Related

Closes #286

- #281 / #284 — added the table.


FrankRay78:

- Bugs or correctness issues

Issue #270, opened 2026-09-05
Add scripts/chain.sh — run the SDLC command chain end to end for one issue


FrankRay78:
## Summary

Add a script that carries one GitHub issue from an unstarted branch to an open pull request by running the SDLC commands in a fixed order — `/build`, `/study`, `/verify`, `/study`, `/raise-pr` — with the script, not a model, deciding whether each stage begins. One invocation, one issue, no prompting in between.

## Motivation

After #257 the chain is three named stages, and #260 adds a fourth thing worth doing twice. Each is unattended on its own; the sequence is not. Today the invoker types four commands, waits for each, reads its closing report, and decides whether to start the next — and the waiting, not the work, is what makes a piece of work take a day instead of an hour.

The obvious shortcut — one agent process told to run all four commands in order — is not equivalent. A prompt listing four steps is a suggestion: a model can reorder them, skip one, decide a stage is unnecessary, or report success for work it did not finish. What makes an ordering deterministic is a process outside the model enforcing it, and refusing to start the next stage until the previous one has said, in a form the script can read, that it succeeded.

Two properties of the chain make that enforcement worth having rather than merely tidy. The stages are expensive — a full run is the better part of an hour of model time — so a stage that runs against unverified work wastes real money and, at the last stage, opens a pull request over it. And the stages are not context-sharing: each runs in its own process, so anything one stage learns that a later stage needs has to be handed over deliberately or it is silently lost.

This issue is also the groundwork for #266. Whether that dispatcher drives this script per item or is merely modelled on it, the question of "what does it take to run these stages back to back without a human" is answered here, once, at a scale that can be run by hand and watched.

## Users & jobs

- **Someone with one issue to build.** They have an issue ready and want to come back to a pull request. They do not want to babysit four commands, and they do want the run to stop rather than push on if something goes wrong partway.
- **Someone diagnosing a run that stopped.** They want to know which stage failed and why, find that stage's output, and pick up the remaining stages by hand without redoing the work that already succeeded.

## Capability

A script that runs the stages in order and gates each one on the last.

**Ordering and gating.** The stages run in a fixed order. Each begins only after the previous has reported success in a form the script reads directly. A stage that does not report success ends the run; nothing downstream executes against unverified work.

**Study in context.** A study pass runs against the work of the stage it follows, with that stage's context available to it, so what it records is grounded in what happened rather than reconstructed afterwards from the diff. This is what #260 asks for when it says to run the command in the same session as the work.

[...]

PR #289, opened 2026-09-12
Review Reason Fact Check

### LINKED ISSUE #287: Check that a /speckit.reviewissue recommendation's Reason holds up against what it cites
## Summary

Every `/speckit.reviewissue` gap ends with a recommendation and a one-sentence *Reason*, and nothing checks that the Reason is true. The command tells the model what a Reason should contain, not that the evidence it cites has to survive being checked against the file it names. Since #281 made reviews short and triageable, an author can accept a recommendation from the table with little reading, so each Reason now carries more weight than it did — and a wrong one settles a real decision quietly.

## Evidence

**A live Reason that does not hold up.** Gap 5 of [the review on #270](https://github.com/FrankRay78/NetPace/issues/270#issuecomment-5646220691) recommends Sonnet for each headless stage because "Sonnet is the harness's standing default for that class of work". That is a claim about how the harness behaves, offered as the evidence for the call. Nothing in this repository names Sonnet: `git grep -i sonnet` over `main` returns no matches, and no file under `.claude/` sets a model. The Reason reads as grounded and is not.

**What the command asks for today.** The [recommendation quality bar](https://github.com/FrankRay78/NetPace/blob/main/.claude/commands/speckit.reviewissue.md#L225-L231) says the Reason "cites the evidence that led you there … not a restatement of the recommendation" — a rule about what a Reason contains. The [pre-post self-check](https://github.com/FrankRay78/NetPace/blob/main/.claude/commands/speckit.reviewissue.md#L223) confirms every gap carries "a recommendation with its one-sentence reason" — a rule that one exists. Neither asks whether it is true.

**The same failure class, recorded elsewhere.** [`docs/study/279.md`](https://github.com/FrankRay78/NetPace/blob/main/docs/study/279.md) records a claim "lifted unverified from the issue's own motivation text" into a command that posts publicly, and an absolute asserted on a two-command sample. [PR #284](https://github.com/FrankRay78/NetPace/pull/284) records three claims in #281's own first draft — a failure mode, a "load-bearing" constraint and a word count — that were false when checked.

**Anticipated, and deliberately left open.** #281's *Open questions*: "The value of a short review rests on each recommendation being a good default and its *Reason* citing evidence that survives being checked against the file it names. Nothing in the command verifies either today, and nothing here adds it. That is a separate issue, and a more consequential one than length." The [CIR records it as a known residual](https://github.com/FrankRay78/NetPace/blob/main/docs/change-intent-records/2026-09-12-triageable-review-comment.md#L39): a shorter review "raises the cost of that gap without addressing it."

## Out of scope

- Judging whether a recommendation is a good default. That stays the author's call when answering; this issue guarantees only that the evidence behind it is true.
- The review's presentation — the at-a-glance table, consequence ordering, the per-gap length bound and the Commentary bar. #281 settled these and the #270 run showed them working.
- Any change to `/speckit.confirmissue`.
- The `|`-escaping gap in a table cell — #286.

### Requirements

- **What the fix guarantees:** Every factual claim in a recommendation's Reason is true when checked against the file, command output or issue text it names. Whether the recommendation is a good default is not checked.
- **Judgement-call Reasons:** A Reason either cites something checkable or states plainly that it is the reviewer's judgement, so the author can tell an evidenced recommendation from an opinion.
- **Unbacked claims:** A claim that fails the check is recast so its Reason states it is the reviewer's judgement, and the Recommendation is kept; a failed check never blocks posting, on a first run or a refine-run edit.

### Technical

- **Who checks:** The reviewer fact-checks its own claims before the review posts; no subagent. (Author redirected from "add the subagent tool to the workflow's `--allowedTools`" — a subagent is overkill for this.)
- **Automated reviews:** In scope — the check applies to reviews posted by [`speckit-reviewissue.yml`](https://github.com/FrankRay78/NetPace/blob/main/.github/workflows/speckit-reviewissue.yml) as well as local runs.
- **Verification call:** No separate verification call, so nothing to batch. (Author redirected from "batch every gap's Reason into a single verification call" — the reviewer checks its own claims.)

## Related

- #281 / PR #284 — made the review triageable; its *Open questions* named this issue.
- [The review on #270](https://github.com/FrankRay78/NetPace/issues/270#issuecomment-5646220691) — the live example above.
- [`docs/study/279.md`](https://github.com/FrankRay78/NetPace/blob/main/docs/study/279.md) — the same failure class inside a command prompt.

_This issue originally collected six follow-ups from the first live triageable review. The other five were dropped: the #270 run showed the table already does its triage job, and the reviews on #285 and #286 (3 gaps each, against 5 on #270) settled the proportionality question. See the edit history for the original._

FrankRay78:
## Why

Every `/speckit.reviewissue` gap ends with a recommendation and a one-sentence *Reason*, and nothing checked that the Reason was true. The review on #270 justified a call with "Sonnet is the harness's standing default", and no file on `main` names Sonnet. Since #281 made reviews short enough to accept from the table, a false Reason can quietly settle a real decision.

## What changes

- Before posting, the reviewer re-checks every factual claim in every Reason against the file, search or issue text it names. A claim about how the repo or harness behaves names no file, so it's checked by searching for where that behaviour would be set, and finding nothing fails it.
- A claim found false is **dropped**, and the Reason is rewritten as `my judgement — <why>` without it. A claim the run can't check may stay, but only inside a judgement Reason. The Recommendation always stays, and a failed check never blocks posting.
- The pre-post self-check, the recommendation quality bar and step 6 (the re-run rules) are aligned with this. A re-run checks only the gap it re-framed.

- **The judgement label never carries a false claim.** The first draft only said to "recast" a failed Reason as judgement, which would have let the #270 claim through as "my judgement — Sonnet is the standing default". The `/verify` review caught it, and the rule now drops the claim.
- **Automated reviews vs local runs.** `speckit-reviewissue.yml` applies this prompt as written, so it's covered with no workflow change. But its tools can't read other issues (no `gh issue view`), so a true Reason citing another issue posts as judgement from the workflow and as evidence from a local run. Allowing `Bash(gh issue view:*)` would close that gap, but it widens the workflow's permissions and wasn't in #287's scope, so it's left for a separate decision.
- **No CI gate.** The check runs over a comment composed at runtime, like the rest of the self-check. It's untested in a live run, so the first review after merge is its first real test.
- **"Judgement call" wording removed.** The prompt no longer calls every recommendation a "judgement call", so "my judgement" marks only a Reason with no evidence behind it.

## Related

Closes #287

- #281 / #284 — made the review triageable; its open questions named this issue.
- The rationale is in CIR `2026-09-12-review-reasons-survive-checking`.


FrankRay78:

- Bugs or correctness issues

FrankRay78:
Pushed `d31a742` — a one-line prompt-text trim, no behaviour change.

The **Recommendation quality bar** paragraph ended with "— or, where there is nothing checkable to cite, says it is your judgement", which restates the opening of **Every Reason survives being checked** directly below it. That paragraph states the rule in full, including the `Reason: my judgement — <why>.` label format, so the clause is dropped.

@claude please review the last commit (`d31a742`).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #290, opened 2026-09-13
Capture Learnings 286 287

FrankRay78:
## Why

Two learnings from building #286 and #287, approved at a `/capture-learnings` checkpoint.

## What changes

- **`/build` stops when local `main` has unpushed commits.** It branches from `origin/main`, so an unpushed commit silently drops out of the new branch. This happened with #281's study record: committed to local `main` but never pushed, it was missing from #286's branch even though #286 cited it as evidence.
- **The grep-after-simplifying memory now covers adding a rule.** A new rule in a command prompt must be reconciled with the older rules it overrides and scoped for every path it applies to (first run, re-run, automated workflow). `speckit.reviewissue.md` missed this on two issues in a row: #286's self-check clause and #287's Reason rule both clashed with the re-run rules in step 6, and `/verify` reviewers caught both.

- **`/study` still doesn't push.** Every pipeline command only commits and `/raise-pr` alone pushes, so the stranded-record problem is fixed at `/build`'s entry instead.
- **The memory was updated in place, not duplicated.** Its title and index line changed to match.

## Related

- #286 / #288, #287 / #289; findings recorded in `docs/study/286.md` and `docs/study/287.md`.


FrankRay78:

- Bugs or correctness issues

Issue #243, opened 2026-08-31
Codebase maintenance backlog


FrankRay78:
Running backlog of codebase maintenance items. Each is independent — pick one up without reading the others. Append further findings as new numbered sections.

---

## 1. Delete 13 orphaned Verify snapshots

- [ ] **1a** — Delete the 13 files listed below
- [ ] **1b** — `dotnet test ./src` green

These `.verified.txt` files in `src/NetPace.Console.Tests/Expectations/` have no owning test method. Verify resolves snapshot filenames at runtime, so nothing reads them and nothing fails — deleting them has no build or test impact.

1. `NetPaceConsoleTests.Should_Continue_Multiple_Speed_Tests_On_Exception.verified.txt`
2. `NetPaceConsoleTests.CSV.Should_Include_IPAddress_And_Hostname_Columns_In_CSV_Output.verified.txt`
3. `NetPaceConsoleTests.CSV.Should_Include_Empty_IPAddress_And_Hostname_In_CSV_When_Device_Identity_Unavailable.verified.txt`
4. `NetPaceConsoleTests.CSV.Should_Include_Error_IPAddress_In_CSV_When_IP_Retrieval_Fails.verified.txt`
5. `NetPaceConsoleTests.CSV.Should_Include_Error_Hostname_In_CSV_When_Hostname_Retrieval_Fails.verified.txt`
6. `NetPaceConsoleTests.CSV.Should_Include_Error_IPAddress_And_Hostname_In_CSV_When_Both_Retrievals_Fail.verified.txt`
7. `NetPaceConsoleTests.Json.Should_Include_IPAddress_And_Hostname_In_Json_Output.verified.txt`
8. `NetPaceConsoleTests.Json.Should_Include_IPv6_In_Json_Output_When_No_IPv4_Available.verified.txt`
9. `NetPaceConsoleTests.Json.Should_Include_Empty_IPAddress_In_Json_Output_When_No_Network_Interfaces.verified.txt`
10. `NetPaceConsoleTests.Json.Should_Include_Empty_Hostname_In_Json_Output_When_Hostname_Resolves_Empty.verified.txt`
11. `NetPaceConsoleTests.Json.Should_Include_Error_IPAddress_In_Json_Output_When_IP_Retrieval_Fails.verified.txt`
12. `NetPaceConsoleTests.Json.Should_Include_Error_Hostname_In_Json_Output_When_Hostname_Retrieval_Fails.verified.txt`
[...]

PR #291, opened 2026-09-14
Comment Residue

### LINKED ISSUE #254: Comment and formatting residue left by Surface Transfer Failures (#222)
Second maintenance list, split from #243 so the two can be picked up independently. Every item here is comment or formatting residue left by the Surface Transfer Failures work (#222) — no behaviour changes, no snapshot changes, and each item is independent of the others.

**Sequencing:** intended to be picked up after #243. Nothing here conflicts with #243, but items 2 and 5 touch files (`SpeedTestCommand.cs`, `JsonConsoleWriter.cs`) that #243 item 4 also edits, so doing #243 first avoids a needless rebase.

---

## 1. Delete the BCL assertion from the XML-doc share-mode test

- [ ] **1a** — Delete the `if (OperatingSystem.IsWindows())` block from `src/NetPace.Core.Tests/ProfileXmlDocTests.cs`
- [ ] **1b** — Trim the now-stale last sentence of the scenario comment
- [ ] **1c** — `dotnet test ./src` green

`LoadXmlDoc_WhileAConcurrentWriterHoldsTheFile_StillLoadsTheDocument` asserts that `XDocument.Load(string)` throws when a writer holds the file. That is .NET's behaviour, not NetPace's, and the constitution's Testing Standards list third-party library behaviour under **Do NOT test**. It also makes the test do different work on Windows than on Unix, which cuts against Principle IV.

The rest of the test is legitimate and stays — it is a real regression test for `LoadXmlDoc` opening with `FileShare.ReadWrite`, and it is what stops the intermittent "The process cannot access the file because it is being used by another process" failure coming back.

Delete these four lines:

```csharp
            if (OperatingSystem.IsWindows())
            {
                Should.Throw<IOException>(() => XDocument.Load(xmlPath));
            }

```

And drop the trailing sentence of the scenario comment, which only described the deleted assertion:

```
        // modes, so the premise is asserted there rather than assumed — on Unix this test still runs
        // and still asserts the outcome, but cannot catch a revert of LoadXmlDoc.
```

so the comment ends at `// access the file because it is being used by another process.` followed by a line reading `// Only Windows enforces share modes, so this test is a genuine regression guard there and vacuous elsewhere.`

**Acceptance:** no `Should.Throw` and no `OperatingSystem.IsWindows()` in `ProfileXmlDocTests.cs`; the test still constructs the competing `FileStream`, still calls `LoadXmlDoc`, and still asserts `doc.Root!.Name.LocalName.ShouldBe("doc")`; `dotnet test ./src` green.

---

## 2. Remove the two comments triplicated across the three run branches

- [ ] **2** — Delete six comment lines from `src/NetPace.Console/Commands/SpeedTestCommand.cs`

`ExecuteAsync` has three near-identical run branches (`--loop`, `--count`, single run) and each carries the same two comments verbatim:

| Line | Comment |
|---|---|
| 67, 110, 151 | `// NetPace's own health, not a network condition: exit non-zero.` |
| 74, 117, 158 | `// This measurement never completed, which --fail-on treats as a failure.` |

Both restate the name of the thing on the very next line — `catch (Exception e) when (IsOperationalFault(e))` then `throw;`, and `if (FailOnRequested(settings)) return 1;`. Both helpers already carry the full explanation in their XML docs at the bottom of the file, and `FailOnRequested`'s `<remarks>` covers the "measurement threw rather than completing" case in more detail than the inline comment does.

Delete all six lines, leaving the blocks as:

```csharp
                    catch (Exception e) when (IsOperationalFault(e))
                    {
                        throw;
                    }
                    catch (Exception e)
                    {
                        WriteError(console, e.Message);

                        if (FailOnRequested(settings)) return 1;
                    }
```

Do not touch the `// User requested cancellation.` or `// Pause before the next speed test.` comments — those predate #222 and match the file's established density.

**Acceptance:** zero matches for `NetPace's own health` and `never completed` under `src/`; the XML docs on `IsOperationalFault` and `FailOnRequested` are unchanged; `dotnet test ./src` green.

---

## 3. Drop the `<remarks>` explaining the C# `required` keyword from `JsonResult`

- [ ] **3** — Delete the four-line `<remarks>` block from `src/NetPace.Console/JsonResult.cs:6-9`

```csharp
/// <remarks>
/// A <c>required</c> property always appears in the output. A nullable one is omitted when its test
/// did not run, so it carries no <c>required</c>: absence is the whole point of it.
/// </remarks>
```

It documents what `required` and `?` mean in C# rather than anything about NetPace's JSON contract. The per-property `<summary>` docs already state which fields are omitted and when — for example `DownloadSpeed` says "or `null` when the download test was not run" — so the class-level remarks add nothing a reader of the properties does not already have.

Keep the `<summary>` on the record itself.

Note for whoever picks this up: #242 is separately deciding `required` on `SpeedTestResult` and `LatencyTestResult`. `JsonResult` is not in #242's scope, so the two do not collide, but if #242 lands first, re-read this block before deleting in case it has been rewritten to say something substantive.

**Acceptance:** `JsonResult` has a `<summary>` and no `<remarks>`; no property doc changes; `dotnet test ./src` green with no snapshot changes.

---

## 4. Remove the duplicated "every request succeeds" comment from `VariableSpeedTester`

- [ ] **4** — Delete two comment lines and their trailing blank lines from `src/NetPace.Core/Clients/Testing/VariableSpeedTester.cs:49` and `:68`

```csharp
                // Every request succeeds; this fake models fluctuating speed, not failure.

```

Both sites sit directly above a `switch` in which every arm already reads `RequestsSucceeded = <n>, RequestsFailed = 0`. The per-call byte/speed comments immediately above are the ones carrying real information and stay.

**Acceptance:** zero matches for `Every request succeeds` under `src/`; the `// Call 1: … Mbps` comment blocks are untouched; `dotnet test ./src` green.

---

## 5. Match the JSON writer's ternaries to the line above them

- [ ] **5** — Collapse two three-line ternaries in `src/NetPace.Console/ConsoleWriters/JsonConsoleWriter.cs:25-31` onto single lines

Currently:

```csharp
        var latencyFormatted = !settings.NoLatency ? $"{fastest.LatencyMilliseconds} ms" : null;
        var downloadFormatted = settings.NoDownload
            ? null
            : downloadResult.GetSpeedString(settings.SpeedUnit, settings.SpeedUnitSystem, settings.SpeedScale);
        var uploadFormatted = settings.NoUpload
            ? null
            : uploadResult.GetSpeedString(settings.SpeedUnit, settings.SpeedUnitSystem, settings.SpeedScale);
```

Three adjacent assignments doing the same job in two different shapes, with the condition inverted between them. Restore the pre-#222 form, which matches `latencyFormatted` and matches how `MinimalConsoleWriter` and `DefaultConsoleWriter` write the same guard:

```csharp
        var latencyFormatted = !settings.NoLatency ? $"{fastest.LatencyMilliseconds} ms" : null;
        var downloadFormatted = !settings.NoDownload ? downloadResult.GetSpeedString(settings.SpeedUnit, settings.SpeedUnitSystem, settings.SpeedScale) : null;
        var uploadFormatted = !settings.NoUpload ? uploadResult.GetSpeedString(settings.SpeedUnit, settings.SpeedUnitSystem, settings.SpeedScale) : null;
```

While here, shorten the two-line comment above them, which restates what `JsonResult`'s property docs say:

```csharp
        // Display speed test result.
```

Not covered by #249 — `dotnet format` does not reflow ternaries, so this will not resolve itself.

**Acceptance:** all three `*Formatted` locals are single-line and use the same `!settings.NoX ? value : null` shape; `dotnet test ./src` green with no JSON snapshot changes.

---

## 6. Remove the redundant `using` from the service-wiring tests

- [ ] **6** — Delete line 1 of `src/NetPace.Console.Tests/NetPaceConsoleTests.ServiceWiring.cs`

```csharp
using NetPace.Console;
```

The file declares `namespace NetPace.Console.Tests`, so `NetPace.Console` is already in scope by namespace walk. No other partial of `NetPaceConsoleTests` carries this using.

**Acceptance:** the file compiles with no `using` directives of its own; `dotnet test ./src` green.

---

## 7. Restore the accurate wording on `TestResult.Output`

- [ ] **7** — Revert one doc line in `src/NetPace.Console.Tests/CommandLineTestHost.cs:66`

```csharp
    /// Gets the output written to the console.
```

back to:

```csharp
    /// Gets the output written to stdout.
```

`#222` reworded this in passing. "stdout" was the more precise term — `CommandLineTestHost` captures standard output specifically, and the CLI does write elsewhere (`--file`), so "the console" is ambiguous about which stream was captured.

**Acceptance:** the doc reads "written to stdout"; no other change in the file; `dotnet test ./src` green.

---

## Notes for anyone extending this list

Append new sections with the next number. Keep each item independent — someone should be able to do item 5 without reading items 1 through 4.

Before adding a formatting item, check whether `dotnet format` (#249) already claims it. Before adding a comment item, check whether the comment is load-bearing: #243 item 6 is the counter-example, where a verbose comment exists specifically to stop a reviewer re-raising a non-bug, and shortening it would be a regression.

FrankRay78:
## Why

The Surface Transfer Failures work (#222) left comment and formatting clutter behind. This removes it. Nothing about how the app behaves changes, and no snapshots change.

## What changes

- Deletes comments that repeat what the code or XML docs already say: the triplicated run-branch comments in `SpeedTestCommand`, the `<remarks>` on `JsonResult` that explain the C# `required` keyword, and the duplicated "every request succeeds" comments in `VariableSpeedTester`.
- Removes the Windows-only `XDocument.Load` throw assertion from `ProfileXmlDocTests`. It tested how .NET behaves, not NetPace. The `FileShare.ReadWrite` regression test is otherwise unchanged.
- Puts each `JsonConsoleWriter` ternary back on one line in the `!settings.NoX ? value : null` form, drops a redundant `using`, and restores "stdout" in the `TestResult.Output` doc.
- Adds `docs/study/254.md`, which records what surprised the build.

- **Item 2's acceptance grep can't come back empty.** `NetPace's own health` is still in the `IsOperationalFault` XML doc, which the issue says to keep as is. `never completed` is still in an unrelated test comment in `NetPaceConsoleTests.Failures.cs`. Only the six lines the issue names were removed.
- **Item 1b wraps differently from the issue's layout.** The replacement sentence was appended to the line above so the comment keeps the block's ~100-column width. The issue had it as its own ~111-character line.
- **Deferred, out of scope:** the XML doc on `LoadXmlDoc` still says `XDocument.Load(string)` "throws instead" without saying that only happens on Windows. Also, the Windows-only assertion came from #237, not #222. That doesn't change what had to be deleted.

## How to verify

- [ ] `dotnet build ./src` has no warnings and `dotnet test ./src` is green, with no `.received.txt` snapshot diffs
- [ ] `grep -rn "Every request succeeds" src/` returns nothing
- [ ] `ProfileXmlDocTests.cs` has no `Should.Throw` and no `OperatingSystem.IsWindows()`

## Related

Closes #254

Follows #222. See also #243, a sibling maintenance list.


FrankRay78:

- Bugs or correctness issues

PR #294, opened 2026-09-23
Make /install-harness-tooling env-agnostic

FrankRay78:
## Why

`/install-harness-tooling` assumed one install route and one machine shape: it deferred its command list to `docs/wsl-claude-sandbox.md`, and gave only the in-session `/plugin` route for context-mode. Both assumptions break on an enterprise-managed Claude account, where `/plugin` can be disabled by managed policy, and on surfaces that never offer it (the VS Code extension), where the symptom is a bare `/plugin isn't available in this environment` with no documented way forward.

## What changes

- The command file now owns its install commands outright rather than pointing at a WSL-specific doc, and states plainly that the tools are Linux/macOS, WSL or not.
- context-mode gains a CLI install route (`claude plugin marketplace add` / `install` / `list`) alongside the in-session route, with the HTTPS marketplace URL used in both because the marketplace default is SSH.
- Adds the recovery path for `claude` not being on `PATH` while Claude Code is plainly running — a `find` that derives the extension's bundled binary rather than a hard-coded path that would rot on every extension update.
- Adds a node >= 22.5.0 prerequisite note for context-mode only, because without node the plugin installs, registers and loads its hooks and skills — so it looks installed — while its MCP server cannot start.

- The two install routes write the same `~/.claude/plugins/` registry, so the result is identical; the CLI route is a fallback for availability, not a different outcome.
- The example resolved path (`~/.vscode-server/extensions/anthropic.claude-code-<version>-<platform>/…`) is deliberately illustrative only — the `find` is the copy-paste form precisely because `<version>` moves.

## How to verify

- [ ] Run `/install-harness-tooling` and confirm the printed commands are self-contained — no need to open `docs/wsl-claude-sandbox.md` to complete an install.
- [ ] Confirm no remaining claim in the file that the WSL doc is the canonical source for these commands.
- [ ] On a box without `claude` on `PATH`, run the `find` snippet and confirm it resolves an executable that answers `--version`.


FrankRay78:

- Bugs or correctness issues

FrankRay78:
@claude Can you please address this PR review comment:

> The one thing I'd want addressed — either in this PR or as an immediate follow-up — is reconciling docs/agentic-workflow-NetPace.md:116 (and optionally docs/wsl-claude-sandbox.md step 7) so they don't keep pointing readers at what is now the less-complete copy of these instructions.

Then commit and push, and tag claude in to re-review the last commit.

FrankRay78:
@claude I propose the following, please assess:

- docs/wsl-claude-sandbox.md - revert the change (it's a historical document, not actively maintained)
- docs/agentic-workflow-NetPace.md - revert then delete lines 116, 117 (this is a sidekick file, it should be additive and not reference it's master)

If you agree with both, then proceed to make the changes and commit/push.

